PR #1, opened 2018-04-24
a shot in the dark

vortexing:
I edited some sidebars and pages to make them start to be similar to the three domains we're envisioning but mainly I just want to get my feet wet with what edits what and waht is connected to what.

@bmcgough I'm basically afraid to break anything you have done here, so I'm going to let you see if we can merge my first pass edits in as a pull request from a branch so that it should be easy to revert anything I did en masse and just ignore it.  It has nothing magical in it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,18 @@
+## How to Contribute:
+Steps to contribute content to the Fred Hutch Wiki Pages
+
+*Assuming:*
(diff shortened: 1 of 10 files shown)


vortexing:
OK, I'm feeling daring and really want to test it!!!

bmcgough:
That's fine - I'm not sure how all the parts relate to each other right now myself. For example, I believe we want to have folders at the root for each domain, but I think the demo had things set as pages. Once the paths are worked out in the links, we'll probably want to remove /pages entirely, or at least remove the domain pages from the folder.

I did get one strange yaml error from one of the sidebar files during the github jekyll build. I could not find the typo and ended up doing a blind copy and replace which fixed it. Maybe a strange control character or something.

PR #15, opened 2018-05-21
Good start to the docker docs

sminot:
We can leave this PR open until the Thursday meeting, so others can push to this branch if they want to add to or edit the Docker markdown


sminot:
@canaantt My feedback would be that calling a Docker image a 'file' is (while technically correct) going to much more confusing to the new user. For example, you can 'run' a Docker image and then interact with it as though it is a complete operating system, which is not the common understanding of what constitutes a 'file'. I'll defer to the group, of course, but I'd advise against that specific terminology.

canaantt:
@sminot I agree with you. It indeed introduces unnecessary confusion to the audiences.

vortexing:
I've gotta say I'm pretty enthused by this.  I know it's basic, but it's great!!!

canaantt:
I got over-enthused...forgot this is not software development...

PR #16, opened 2018-05-22
Initial description of AWS Batch

sminot:


dtenenba:
Image name should be `ubuntu:16.04` instead of `ubuntu/16:04`

sminot:
Thanks @dtenenba

sminot:
Hopefully that helped address your comment, @vortexing

vortexing:
Yuppers.  I also added some links and super basic info.  Perhaps this page should be an AWS page?  I could see how we could explain the various ways FH researchers can use AWS products for their work.  Maybe that'd be too big.  But into to getting/using an S3 bucket (regardless of your intent to use AWS Batch), is something important.  So is using S3 so that you could also use Globus instead of AWS Batch if your intent is slightly different (more of the slow-cooker approach - set it and forget it, than AWS Batch where you're still using custom tools, etc and setting up a Docker image, etc).

PR #18, opened 2018-05-23
Can I push to a merged branch?

vortexing:
Let's get a few more comments and input from novices or SciComp about this before we merge it into the master again!


vortexing:
Would y'all look at my novice-edits and see if I made it clearer or not?  Then let's merge it in if they are ok.

PR #33, opened 2018-06-14
first stab for scratch fs

dirkpetersen:
pasted in section from sharepoint and added introduction


vortexing:
@dirkpetersen are you thinking you're done here for a while?  Should we merge what we collectively have so we can get it live?

PR #37, opened 2018-06-20
Scicomp shiny

canaantt:
@vortexing @robert-mcdermott


vortexing:
This is still a work in progress and I think we should continue to work on it before we merge it in.  Let's work on this another week with a goal to put something final in by NEXT Thursday.   This is in my court and I will request a review from @canaantt and @robert-mcdermott to make sure I didn't over-idiot this.

PR #39, opened 2018-06-26
First cut at the storage overview

atombaby:
Added significants amount of content to the overview.  There's still a couple of pieces of information required (specifically on whether the storage collaboration items are OK for PHI/restricted data) but those shouldn't change the structure significantly.

I'd like everyone to have a look, make sure I haven't got things too wrong.  I've oversimplified in some places, but I think there's already too much detail.


atombaby:
Ok, @dtenenba I've added some stuff which I hope addresses your concerns.  For the time being, I'm leaving S3 as being only for collaboration purposes, with the caveat that the section on cloud computing will have a longer discussion of its capabilities.

Let me know if thats sufficient.

dtenenba:
Looks good to me!

atombaby:
Ok- done.

PR #44, opened 2018-07-05
Matlab

fizwit:
Instructions for using Matlab on gizmo


vortexing:
@fizwit Is there someone on campus that uses Matlab a lot still that I could ask for input on Matlab docs?  I can certainly review as a novice (haven't used Matlab since grad school) but I'd certainly be willing to go find a more expert reviewer to help out!

PR #45, opened 2018-07-10
Feature/using s3 from r

dtenenba:
Seeing feedback.

### DIFF
--- a/_computing/resser_s3_from_R.md
+++ b/_computing/resser_s3_from_R.md
@@ -1,0 +1,154 @@
+---
+title: Using S3 from R
+last_modified_at: 2018-07-09
+---
+
+# Using S3 from R
+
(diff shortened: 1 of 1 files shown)


_computing/resser_s3_from_R.md
```diff
@@ -0,0 +1,154 @@
...
+```R
+library(aws.s3)
+```
+
+**NOTE:** The example fragments from this point on assume you are in an `R` session with `aws.s3` loaded.
+
+### List all buckets
+
+```R
+bucketlist()
```
atombaby: Would it make sense to indicate handling returned values?  Like:

    blist <- bucketlist()

dtenenba: Good idea, done.

PR #46, opened 2018-07-12
X11 for Mac

fizwit:
XQuartz and terminal usage for Users


vortexing:
@fizwit So, perhaps this is just my naivete but because you forked this repo first rather than just creating a branch from it, I can't go in there and move the file  to the right place for you.  SO, would you please be so kind as to change the filename now to this:

_computing/user_MacX11.mc > _computing/linux_MaxX11.md

And then I can put more edits into the document and such once it's merged into the master and I can make a new X11 branch.

vortexing:
Replaced by pull request #59

PR #75, opened 2018-08-03
Proposed addition of pages for HPC on prem/cloud

vortexing:
I added a page for cloud computing and for rhino/gizmo, b/c these two pages are ones that many researchers have needed.  Maybe one will be fairly small for now (cloud), but the rhino page could get filled ASAP.


atombaby:
I'd like to combine beagle with rhino and gizmo in the same section- it's all the same "style" of computing (i.e. shell based/command-line/batch).  Even though beagle is in EC2, from the end-user perspective it's nearly no difference.

PR #81, opened 2018-08-23
Cluster rhino gizmo

atombaby:
Added introductory information on interactive compute options.  @bmcgough @fizwit can you review what I've written about notebooks and rstudio?


atombaby:
Note removal of `hpc.md`.  I did take a quick look and I don't see it referenced anywhere.

PR #89, opened 2018-08-29
updated Jupyter sections

vortexing:


vortexing:
I added more basic stuff to this and some links but now it won't let me review it.  Someone else should though.

PR #96, opened 2018-09-06
changes to object storage, economy local and economy cloud 

dirkpetersen:
also  corrected some errors in the overview.


vortexing:
I have some feelings about where S3 belongs in this whole thing as well.  But perhaps like you say, that's v2.0.

PR #97, opened 2018-09-06
build locally and check for broken links

### LINKED ISSUE #94: Ongoing link checking
An issue we will need to address in the future is to check to make sure all of these links we are putting in go somewhere valid.  Maybe it's the wrong place (thus a human has to check these), but at least no 404's, etc.  Is it possible to set up some CI that will address this?

dtenenba:
This would close #94 .

I managed to get the site built locally. I have added some documentation on how it is done in the README.

I could not get link checking to work inside jekyll, but I did integrate the repo with (hosted) CircleCI.

It now checks for broken links (and finds 149 of them!).

See:

https://circleci.com/gh/FredHutch/wiki/6

...for the build report which lists all the broken links.

Note that one side effect of adding CircleCI is that all builds on other branches will fail the CircleCI build because those other branches do not have a circleci config file so Circle does not know what to do.

After merging my changes, there will be a circleci config file, however, builds will still fail until all those broken links are fixed. 

Note that the CircleCI build failing has NO effect on the Jekyll build and propagation to the github pages site. The jekyll build should work no matter what.

If desired, we can tweak things so that only certain branches are built by CircleCI. Maybe it would make sense to only build `master`.


dtenenba:
Note also that the link checker does more than check for broken links. it's super finicky and checks for things like missing alt attributes on image tags (which are important for people using screen readers). However, it's highly configurable and can be told to ignore anything that's not an actual broken link. See https://github.com/gjtorikian/html-proofer#configuration

dtenenba:
You can also put an attribute inside links that you do not want to check, and the checker will ignore them. This might be appropriate for links to hutch-only resources (I see one in the report to galaxy.fredhutch.org), which the hosted CircleCI can't see. I chose to use the hosted CircleCI rather than the on-prem one because that way the build reports would be visible to all (our on-prem CircleCI has a license limiting the number of users). 

Note that you can run the link checker yourself (based on instructions in the README). I just did this and found 139 failures, fewer than the 149 found by CircleCI, so the difference must be attributable to my being able to access on-campus-only resources.

vortexing:
Whew.  DANG!  OK, this seems great.  These are my thoughts:
- I think it'd be best first off to have CI pass if all the links are fine (and not have it get hung up on alt attributes for now b/c I wont' have time to fix all of that for a while).  Then once we usually have it pass, then we can add in more refined checks that are important but not mission critical.

- Will I ever be able to fix everything so that it passes if many of the links go into Centernet?

vortexing:
Also, on a Mac, I don't have permissions to install bundler locally ("You don't have write permissions for the /Library/Ruby/Gems/2.3.0 directory").

dtenenba:
@vortexing try adding "sudo" before the 'gem install bundler' command. If that works I will update the docs.

I will fix it so it doesn't complain about attributes. 

As for the center-only resources, I need to think on that.

dtenenba:
OK I have gotten it down to 2 broken links which are true positives. They are listed at 
https://circleci.com/gh/FredHutch/wiki/31

and here:

```
- ./_site/bioinformatics/inf_index/index.html
  *  internally linking to /infdemos/, which does not exist (line 453)
     <a href="/infdemos/">Research Topic Articles</a>
- ./_site/generation/gen_index/index.html
  *  internally linking to /generation/study_statistics/, which does not exist (line 470)
     <a href="/generation/study_statistics/">Statistics</a>
```

If you (@vortexing) or anyone else have ideas on how to fix them let me know and I will do so, or we can just let them be and fix them in master after we merge this. 

Note that internal-only URLs that CircleCI can't check are listed in the file Rakefile.
We might need to modify this if we add additional ones.

vortexing:
sudo works.

For internal links I would like to still check them somehow, even if it's just generating a list of them that some human needs to click on on campus for now.  I'm hesitant to just ignore them all and hope they are ok.

vortexing:
Fixed those two broken links.

dtenenba:
OK, you can now test locally with `rake testlocal` instead of `rake test` and it will include all internal links that aren't password protected. This is documented in the readme.
Also, your (@vortexing ) latest commit introduced some more missing links....

vortexing:
aaarrgghhh.  The missing link!!!

vortexing:
OK, so after all this, I should merge this pull request right so that it can start checking the master branch where I just committed my edits to fix links b/c the master currently doesn't have a config file.  Am I understanding that this is ready to merge and then perhaps future tweaks occur as we get used to it?

dtenenba:
Yes, that's correct.
Dan

dtenenba:
It strikes me that this link checking is of limited utility since not everyone will notice that the check has failed. I am  the only one who  gets emails about it, and aside from that the only way to know if the build has passed is to proactively look at the commits page (https://github.com/FredHutch/wiki/commits/master) and see if there is a green checkmark or a red X. 

We can add a slack notification that will tell us if the build (link checking) fails, or when it's working again, and which commits triggered either state. The only questions are: which slack team should the messages go to (SciComp or FHBig) and which channel should they go to, an existing one or a new one? I would vote for the FHBig slack team as that is open to more of our contributors, and suggest creating a new channel for circleci notifications, and then asking all contributors to subscribe to that channel. If that sounds ok (@vortexing ?) I may need some help from an admin of that slack team (you?) in setting this up.

vortexing:
Can you have the link fail email sent to `sciwiki`?  Do you want to be on that listserv too?   OR Slack is fine too (in that case, yes, to the FHBig is fine, and the #wikicontributors channel gets GitHub updates already so perhaps consolidate to that channel?).  I would consider link breaking to be something of an admin thing to address though so I'm not sure that is really something the entire community really needs to know about.  But that's just my 2 cents.

dtenenba:
In that case maybe the notifications should go to their own channel and therefore only go to those who are actually interested in them. I don't know who's on the sciwiki list or what it's for and I am not even sure if I can control who gets emails, it might just be hardcoded to the email that's tied to my github account. 

There are false positives (sometimes a legit website is just slow for a second and times out) and I don't want to spam people. So maybe create a new channel like #wiki-link-checks or something?

Also, at the moment any branch besides master will fail since master is the only one with the circleci stuff set up. I should probably change things so that only pushes to master trigger the circleci build, that will also cut down on spam.

vortexing:
Actually, I don't know how to get Slack set up to get any notifications besides issues, commits, pulls and merges. Those are all of the types of notifications the app seems to have.

dtenenba:
I will coordinate with you to set this up. I think that CircleCi does the notification, and slack just receives it. But I will need some sort of access token that you will need to provide. I am working on some other stuff this morning but will get in touch later about this.

Issue #130, opened 2018-10-04
Plan out the Scientific Computing section in Computing


vortexing:
**Proposed Domain**
Computing

**Content Summary**
http://sciwiki.fredhutch.org/computing/cluster_overview/
Let's write a summary of what we intend to put in each page so we can better organize and reduce duplication of content.

PR #131, opened 2018-10-08
Refactor of scicomp landing page

atombaby:
Getting started on #130.  The landing page has been changed to`_computing/scicomp_overview.md` and the contents redone from `cluster_overview.md`.  The hope is to make a bit more of a narrative flow between sections and add more detail about what the reader will find at the link.  This should allow readers to quickly jump to the area of interest.


bmcgough:
I think that comp_index _itself_ should be the SciComp overview, and not an additional page. What do you think?

I'm in to fewer pages.

atombaby:
I think I thought this was the sidebar.

vortexing:
hmmmm, I too am confused. 

comp_index is the first landing page you come to when you click on Computing.  This should have an overview of each of your main sections (Access and Credentials/Storage/Scicomputing/Software, Resources, Updates).

THEN within each of those four sections, you have a {section}_overview page that should be the overview for the section only.  Then each page in the section is its own markdown with {section} in front of it.

vortexing:
It might perhaps be best to write this out, decide on names for your sections, update the file names so they make sense {section}_overview, {section}_topic1, and then stop changing names.  I see a lot of evolution and name changes and that really makes git/jekyll/Amy confused.

bmcgough:
We will talk about this tomorrow at our meeting and try to come to agreement on the overview page, and sub sections.

atombaby:
We're working on a different approach.

PR #134, opened 2018-10-11
Minor changes after team meeting

bmcgough:


vortexing:
Note:  After a convo a while back, the Resource Libraries in all the domains were less than obvious as to their location.  So all three domains now have  Resource Libraries (generation recently got one), and I moved them all up to the very top of the sidebars.  This way they are in the same place regardless of which domain you're in.  You may want to remove the replaced compdemos link at the bottom of this sidebar if you'd like to keep your domain in line with the other two domains, OR remove the Resource Library link at the top of your sidebar and go rogue.

bmcgough:
I will adjust the link. Doesn't hurt to have both, right? I like words like cookbook, recipe, howto, or demo, so I'd like to include that in our sidebar somewhere.

Issue #133, opened 2018-10-10
Add How to *Receive* data from Caisis when it contains PHI


vortexing:
**Proposed Domain**
Generation, specifically: http://sciwiki.fredhutch.org/generation/clsp_clinicalCov/ and a bit of Computing

**Content Summary** 
- The description of how to request clinical data is there, BUT is it ok to email MRN's to someone to specify which patients you want?  

- When the data is given to the requestor, if it has PHI in it or not, how should they plan on receiving that data?  Via Secure storage (http://sciwiki.fredhutch.org/computing/store_posix/#secure-file)?? What other storage/transfer options exist and when should you use them with respect to Caisis data? @Rachierizz (I can't tag Emily who I have a sneaking suspicion is the person who'd know this, but maybe you know who this should be?)

- If someone does not appear to have access to the relevant PI's secure storage, how does one gain access?  @dirkpetersen How does one make sure that the people who have access to this secure storage location are the ones that can see the Caisis data of interest?  

- Has Center IT/HDC come up with the data transfer/storage/permissioning solution for this situation yet?  

- Does it make sense to give the Caisis data back to investigators by spinning up a REDCap project to give it to them and thus avoid all of this mess? @plitwin @Rachierizz  Was this the actual plan?

**Local Content Expert(s)**
@TimMonahan is a better source of what is/was confusing and what guidance he may have wanted during this process.   I'm simply a conduit and I know nothing about any of this.

vortexing:
@esilgard is in!  Let's figure out what tidbits we can put into the docs here or elsewhere and link to those docs in the Wiki.  Some things aren't worked out yet, and that's ok, but let's write down what the plan is where we do have one!! Can we all a little in here so no one person has to do all the lifting?

vortexing:
Y'all I made a branch for @sgglick 's input she was planning to contribute today, and anything we can get from the players on this issue thread, would be great to put all in this file (_generation/clsp_clinicalCov.md) in this branch (called Caisis-request-receive):
https://github.com/FredHutch/wiki/blob/Caisis-request-receive/_generation/clsp_clinicalCov.md

Let's pile all of our edits there and then can do a pull request when everyone has weighed in!

vortexing:
Tidbits of inspiration:
[...]

PR #145, opened 2018-10-19
Caisis/HDC Data Request process

vortexing:
This branch needs to be addressed.  There was some discussion outside of GitHub but it highlighted some issues and some guidance from those who are owning this process would be great as it doesn't seem to be documented elsewhere in an accessible place.  If it IS documented elsewhere, someone please put in the link!!

Issues appear to still include:
- What information is provided on the submission form only that if a requester forgets, they need to be reminded of (could that content live in SciWiki if it doesn't already have a home, in which case put in the link)?
- What are the modes of RECEIVING data that has identifiers in it, and how are future updates to that data handled (e.g., if more patients participate in a study then more up to date data would be desired on an ongoing basis)?
- What guidance is there for when MRNs need to be sent by a requester (how should this be done?!?!?!)?

Other thoughts @TimMonahan and @Rachierizz  and @esilgard ?  Y'all do this thing.


vortexing:
Addresses (hopefully) issue #133

TimMonahan:
re: the issues noted above by vortexing...
'Info on submission form' is it possible to return a pdf of the submission to the requester? Something like that would be easy for the requester to save and refer to later.
'Modes of receiving' in another system I have the ability to use R and connect directly to the server and get data from a sql view. This method seems to prevent any proliferation of files and leaves me more in control how current the data is.
'guidance for sending MRNs' having a common folder available to users (and a link to the folder sent to the requester) seems the best solution to this and also a good optional solution to the previous issue as well. idk. That is my 2 cents.

sgglick:
Question: 'Info on submission form' is it possible to return a pdf of the submission to the requester? Something like that would be easy for the requester to save and refer to later. Yes, REDCap has a way to capture a completed submission in PDF.  I could add a question asking if the requester would like to receive such PDF, and then I can manually send it.  Will check REDCap to see if there is an automated response which can include such a PDF.
Question: Modes of receiving a file.  Ok, working on this. I think the ability for HDC to direct extracted data directly to a server is a great idea.  I have to talk to the engineers.  HDC has embarked on a longer tem solution for sharing data; how to improve the release the responses to a data request in the near future is something we can work on. 
MRN lists transfer from requester to HDC - the submission process now keeps the MRN listing within the automated process for completing the request.  An alternative shared folder with access via a link, which can be placed in the submission form will strongly be considered.  Thanks for the 2 cents.

vortexing:
(If you guys comment that this is ok, then I will use Admin Privileges to get it merged!!)

vortexing:
Addresses issue #133

PR #148, opened 2018-10-22
Human Specimen's Study Guidance: 3 use cases

sgglick:
First Draft of new section


vortexing:
@sgglick would you mind looking at this table I made to hopefully make this simpler?  I noticed that a lot of the steps were actually the same, it's just that some steps you can skip depending on the scenario.  It's useful to know what you DON'T have to do as much as it is what you DO.

vortexing:
It'd also be nice to link to the IRO form OR the IRO office (so if their form link breaks hopefully people will still be able to search their forms for the title of the form we provide) in where references to specific Fred Hutch forms exist.

PR #154, opened 2018-10-29
Formatting repair on access_credentials.md

vortexing:
Minimal edits, no content change, just formatting

### DIFF
--- a/_computing/access_credentials.md
+++ b/_computing/access_credentials.md
@@ -5,8 +5,6 @@ main_authors:
 primary_reviewers:
 ---
 
->Note:  This section is still under construction.  If you have any input for what you'd like to see addressed here, please email your suggestions to `sciwiki`.
-
 This section describes how to get access and credentials to computing systems on campus and in the cloud.
 
 ## HutchNet ID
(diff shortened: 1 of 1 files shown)


dtenenba:
You should specify how to contact scicomp. If a person is new enough to not know how to get on a rhino, then they will not know how to contact scicomp. So it should explicitly say "email scicomp@fredhutch.org".

dtenenba:
@vortexing  You are right and I'd forgotten about this. 
Maybe we could just say "email scicomp@" and leave out the domain name. 
When I type "scicomp@" into my mail client, an autocompletion suggestion comes up with the full address. So that's probably enough.

vortexing:
Will writing username@ actually avoid the robots???  I like that convention too.  It's totally doable!!  Maybe we should try to change over?

dtenenba:
I'm worried that `address@` will be too easy for spammers to harvest. All they have to do is add the domain name that the page is on. I think we've hit the sweet spot the way it is.

vortexing:
OK sounds good.  Let's just fumble through here and then if we can add some sort of info for people somehow via a hover or something then we will!!

dtenenba:
Any content of a hover/tooltip can also be harvested. ;-(

Issue #160, opened 2018-10-30
Make wiki conventions more obvious to readers


vortexing:
There are some communication with readers issues that I'd like to ponder/improve as we go:

- [ ] We have used the convention of putting usernames in `backticks` but we need to tell people more clearly now that that means you just add the @domain afterward

- [x] Make it more obvious that they can edit any page by clicking on the "Edit This Page" link - perhaps bring it up to the top of the pages?  or in the TOC on each page?

- [x] Make it more obvious how they can comment on or ask for new content by clicking on teh "Send us Comments" link - perhaps also bringing that along to the top OR put it in the TOC on each page?

vortexing:
@ptvan and @jdberndt What do you think about me putting in a small chunk of html into the top of every content page just below the front matter so that all our pages would look like this drafts template:

https://sciwiki.fredhutch.org/drafts/contentTemplate/

Would that help with the second and third issues in this issue at all even if it's just a first pass?  Or should I wait for something more comprehensive?

jdberndt:
Looks good to me other than I would prefer there be a little more whitespace after the word "Github"?

vortexing:
Me too on the whitespace!!!  I tried a couple things but I forget my html!  

If I can fix the whitespace then I will endeavor to move this into the main site.

What do you guys think about changing the username convention from `username` to `username@`?  Does that make it more clear that it's the beginning of an email?  if you type that into your mail client, it'll likely autocorrect to the domain.

ptvan:
I'm good with `username@`

vortexing:
From another thread, it seems that bots can pick up the `username@` as well and knowing the domain of the site can slice/dice/spam. :(  So I think for now `username` will stay until any of us have another Big Idea.  I'll leave this issue open for now too.

vortexing:
Let's give up on this for a while.

PR #163, opened 2018-10-30
Make edit this page/comment link show up at top and bottom of pages

vortexing:
This addresses issue #160 to some degree.

### DIFF
--- a/_drafts/contentTemplate.md
+++ b/_drafts/contentTemplate.md
@@ -2,8 +2,7 @@
 title: Topic Title
 last_modified_at: 2018-06-06
 ---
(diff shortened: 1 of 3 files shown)


vortexing:
Guys, basically I took the other edits on that Drafts page and actually put it inside the html parts of the site so that all the pages will automatically look like what I showed you without us having to edit every.single.page.

vortexing:
screen shot 2018-10-30 at 1 13 05 pm
Looks like this on my local rendered copy.

vortexing:
hahaha.  Well we're using the theme as a remote theme, so I had to make a duplicate single.html file in our repo that would override the one in the remote theme repo.  But it's consistent across all pages and I didn't have to copy/paste into every single one!  WOOT!!!

PR #166, opened 2018-10-31
Update README.md

ptvan:
The page is a bit long, I think a preview in the form of a TOC would be nice. Also, we tell contributors to use TOCs in their articles, so a bit ironic that this page itself doesn't have one :). Also, cleaned up some more text, and added free screencast tools for Linux and OSX


dtenenba:
I think Jekyll provides a way to add automatically generated TOCs. These are less work and more accurate as they are rebuilt whenever a header changes. I'll look into how to do this.

vortexing:
Jekyll makes all the TOC's on the website but Phu was editing the readme in the repo.  So no Jekyll.

ptvan:
Good to know, thanks @dtenenba !

dtenenba:
Duh, got it, thanks for the explanation @vortexing.

Issue #168, opened 2018-10-31
Privacy and Security page for Human subjects as missing link


vortexing:
**URL of page(s) of concern**
https://sciwiki.fredhutch.org/generation/human_privacySecurity/

**Describe problem**
"Data/Information should be handled according to the FH Information Classification Standard, which requires security and handling standards commensurate with the risk of information mishandling. An easy to read chart is located HERE. "

HERE does not have a link!  Where is this pointing to?

sgglick:
Such a chart does not exist on FH CenterNet. Moving communication to email

sgglick:
Will insert a link to the standard as written.  There is no quick view/graphic representing the standard.

sgglick:
There should eventually be an easy to read chart on the ISO site to which to link.  One does not exist now.

vortexing:
Did this!

Issue #60, opened 2018-07-19
Create a quick user facing doc on data classification 


dirkpetersen:
We are using terms such as  "protected data"  in the wiki. The hutch has now a data classification standard 

https://centernet.fredhutch.org/cn/u/hdc/_jcr_content/leftParsys/download_0/file.res/Fred-Hutch-Information-Classification-and-Handling-Standard-FINAL.pdf

Can we motivate HDC/ISO to write a quick summary for the Wiki

vortexing:
Sounds great.  I think that a document about this issue would be great to add to this existing page:
http://sciwiki.fredhutch.org/generation/human_privacySecurity/

That is where we have started putting information about how to think about privacy and security when human data is included.

vortexing:
@dirkpetersen do you have any updates on getting someone to contribute to this?  Have you talked to @sgglick about it?  Is that something more appropriate for someone else?

sgglick:
Dirk did talk to me and I did create one and sent it to ISO to review a few days ago.  I have attached a chart you can use (ISO has not formally approved, so we may need to revise once they do).  The main issue to be aware of is that standard demands functions not currently available to  most at FH: encryption at rest and encrypted email.  There is a notation about this in the chart; it was decided to postpone the effectiveness date of the encryption requirement.

I have also provided a PII vs PHI chart and the 19 elements of de-identification (within same excel file).  Finally I have posted an in-depth de-identification guide for structured, NON-Genomic data.

Let me know…
Susan

Research Compliance Specialist / Hutch Data Commonwealth / 206.667.2507 / sgglick@fredhutch.org<mailto:sgglick@fredhutch.org> / Fred Hutch / Cures Start Here

vortexing:
Will follow up with Susan via email.

sgglick:
Dirk did talk to me and I did create one and sent it to ISO to review a few days ago.  I have attached a chart you can use (ISO has not formally approved, so we may need to revise once they do).  The main issue to be aware of is that standard demands functions not currently available to  most at FH: encryption at rest and encrypted email.  There is a notation about this in the chart; it was decided to postpone the effectiveness date of the encryption requirement.

I have also provided a PII vs PHI chart and the 19 elements of de-identification (within same excel file).  Finally I have posted an in-depth de-identification guide for structured, NON-Genomic data.

Let me know...
Susan

[...]

PR #171, opened 2018-10-31
Added data classification table

vortexing:
Addresses issues #168 and #60.


vortexing:
The table is messy but it renders ok.  That's all the time I have for this now. Sorry.

Issue #85, opened 2018-08-26
AWS Batch Doc integration and links


vortexing:
Is this repo going to be finished for it's docs?
https://fredhutch.github.io/aws-batch-at-hutch-docs/

Is this related?
https://github.com/FredHutch/aws-batch-helpers

Is this something that could be a nice demo in this site as well?  
https://github.com/FredHutch/sra-pipeline

What makes the most sense to link all of these pots of docs with what @sminot  wrote about Docker, 
(http://sciwiki.fredhutch.org/bioinformatics/compute_docker/) and AWS Batch, (http://sciwiki.fredhutch.org/bioinformatics/compute_accessing/)?

sminot:
1. I'll leave the question of whether we can finalize https://fredhutch.github.io/aws-batch-at-hutch-docs/ to @dtenenba as a question of what SciComp's plan for AWS Batch is at FH.

  2. https://github.com/FredHutch/aws-batch-helpers is a tool that I use for myself, but I'm not really planning to try to popularize it as a general tool, since it requires a fair amount of setup that would take a lot to train people on. I think the overhead is probably too high to try to push out yet another pipeline tool...

3) I _think_ what makes the most sense to link everything together will be to write an intro to *workflow managers* at FH, which will use Docker images and deploy them in the cloud. That said, I don't think we're quite in the position yet to take that on.

Thoughts from anyone else?

vortexing:
@dtenenba What do you think about this?  Did we come to a conclusion about AWS BAtch docs?  

The pages taht currently mention it are:
http://sciwiki.fredhutch.org/compdemos/batchComputing/
http://sciwiki.fredhutch.org/compdemos/awsBatch/
http://sciwiki.fredhutch.org/computing/cluster_cloudCompute/

We should consolidate.

dtenenba:
We are at the beginning of a cloud redesign. All of this is subject to change. That said, I should maybe document the current process, as it will be a while before the redesign is done.

vortexing:
So perhaps we should move all references into one compdemo to hold the content until it's ready?  Then remove text in in the other pages?

vortexing:
Proposal:

- [ ] Add AWS Batch demo content from Sam to this page: https://sciwiki.fredhutch.org/computing/cluster_cloudCompute/

- [ ] Move in content from Dan's repo:  https://fredhutch.github.io/aws-batch-at-hutch-docs/

- [ ] Consider if a demo entry needs to exist still but with different content

[...]

PR #174, opened 2018-11-01
Combined demo, and dtenenbaum's other doc's site markdown

vortexing:
Addresses issue #85.  

Dan:  I did not copy over your first example in the other repo's markdown.  If there are examples of using AWS Batch, let's put those in the compdemos folder eh?


vortexing:
Let's leave this for @dtenenba to review and merge when he gets a moment.  ;)  I had to take out the CODEOWNERS file until we can figure out what's going on with notifications so it's not going to require us to get a review from him/SciComp.

PR #184, opened 2018-11-08
jupyter lab start instructions

fizwit:
Simplify the start instructions for jupyter notebooks on rhino's.
the option --no-browser will pick an unused port above 8888 automatically.


vortexing:
Closes #184

PR #185, opened 2018-11-13
Update compute_uis.md

ptvan:
Added directions for running RStudio on rhino on all 3 OS's. Probably needs better/tighter integration with the "Computing Resource Access Methods" page... Also, do we want to combine and re-organize the Pro's and Con's for running Jupyter and RStudio locally vs. remotely ? The Pro's and Con's are essentially the same regardless of the tool being addressed...


vortexing:
Hey, @atombaby and @bmcgough  do you have any particular input on this topic?  

@ptvan Let's add the overview section at the top to describe how you decide to run it locally vs on Rhino.

Issue #169, opened 2018-10-31
Summary of where to store different types of data


ptvan:
**Proposed Domain**
Computing

**Content Summary**
We've been getting multiple questions along the lines of "where can I store my data on rhino if it contains personal info?" and "how much does that cost? how often is it backed up" etc... Perhaps a table showing data type, location and basic policies (including cost) ?

**Local Content Expert(s)**
@FredHutch/scicomp

ptvan:
Closed since @vortexing's reorgs and new content addressed this.

PR #186, opened 2018-11-14
Storage service

dirkpetersen:
Added storage FAQ

How much does storage cost ?
How can I see how much data I am charged for ?
Where can I store protected health information (PHI) data ?


vortexing:
The overview page is not the place for this content.  Can you move what you added over to this page:
https://sciwiki.fredhutch.org/computing/store_posix/

The Markdown is:
/_computing/store_posix.md

vortexing:
@ptvan Do you have any insight as to how to include and expand this info to address issue #169 .

vortexing:
OK Let's merge this in.  I moved Dirk's info around and I think it'd good to go.

Issue #178, opened 2018-11-05
Refine BP for Research Data Management section


vortexing:
**Proposed Domain**
Bioinformatics
https://sciwiki.fredhutch.org/bioinformatics/dm_overview/

**Content Summary**
- Overview of where to store what type of data, for a researcher, with links to details in Computing
- Overhaul data ingestion (link to Collaborative Data Storage in Computing) and more specifically, clean up/flesh out use of Public Data Sets
- Edit Archive page
- Add a Using Scratch in your Analyses page - with links to scratch info in Computing but more of a user guide with some use cases defined.  

**Local Content Expert(s)**
Anyone really

vortexing:
Now irrelevant and done!

PR #191, opened 2018-11-26
Data storage movement between Bioinformatics and Computing

vortexing:
I did a first pass on editing the best practices docs in Bioinformatics and tried to do more contextualizing and linking now that computing docs are up and linkable!!  This likely is a work in progress pull request, so weigh in!


vortexing:
Somewhat address Issue #178.  Is it enough for now or should we leave this here until later?

PR #202, opened 2018-12-06
Updated per issue 197

bmcgough:


ptvan:
Content-wise this looks great, though in my preview of the changes the table is not rendering. IIRC @vortexing also saw this...

PR #203, opened 2018-12-10
how to use snapshots

bmcgough:


ptvan:
This is excellent information ! @vortexing Where should we reference it ? Maybe https://sciwiki.fredhutch.org/computing/cluster_rhinoGizmo/ ?

vortexing:
I'm curious to know what storage services have this feature.  I would suggest linking to this HOWTO from any of the Storage pages for which it applies (so only File Storage b/c not object nor collaborative nor scratch right?).  And if you want to toss a link and a blurb in rhinoGizmo, go for it!

bmcgough:
This is one that I think people will most likely find by searching "undelete" and "recover". We can link it where appropriate.

All home directories and shared "fast" folders mounted on linux systems here have this feature.

PR #209, opened 2018-12-14
Cluster node table

bmcgough:
My suggestion is we merge this commit, and then we can move the tables elsewhere later.

Note that the tables are _extremely_ sensitive to whitespace!


vortexing:
It seems only the auto-generating tables are.  The other tables elsewhere in the site seem to not have the same trouble.

Issue #205, opened 2018-12-12
best practices - how to avoid adding AWS credentials to git repos


dtenenba:
**Proposed Domain**
Specify where should the new content should go (Generation, Bioinformatics, Computing, Resource Libraries)

Computing

**Content Summary**
Summarize the topic you'd like to be expanded upon or added.  

Best practices around AWS credentials and version control. 

NEVER put your AWS credentials into git. Ideally these credentials should not be in any file in a local clone of a repository, even if it's a file that does not get committed to the repo.

AWS Batch, Nextflow, and Cromwell each have their own idiosyncrasies around this and each should be discussed.

We have had a couple instances of credentials being committed to a repo and we really need to up our game in terms of getting out the word about what not to do.

**Local Content Expert(s)**
Suggest any Fred Hutch based experts who we might ask to contribute.

I am probably the person to do this, or at least contribute to it, mostly filing this as a reminder to write this up.....

vortexing:
I put a brief big picture mention into the Bioinformatics Github page (https://sciwiki.fredhutch.org/bioinformatics/compute_github/) and then linked to the Computing/access_credentials page.  I added a small section (a duplicate of what was in the other page) if you want to add more detailed guidance there @dtenenba .  

Current branch these edits are in is githubCredentials.  

Here is the link to the actual markdown IN that branch:
https://github.com/FredHutch/wiki/blob/githubCredentials/_computing/access_credentials.md

vortexing:
this was addressed.

PR #212, opened 2018-12-18
GitHub credentials

vortexing:
@dtenenba maybe take a stab at adding some details to the page in computing to close issue #205


vortexing:
Details for guidance from SciComp can go into `_computing/access_credentials.md` so that you all can manage those in the future.  The Github page over in Bioinformatics brings up the issue but then links to this for specifics.  @dtenenba

PR #217, opened 2019-01-03
all info about toolbox.fhcrc.org

dirkpetersen:
toolbox is the API that HDC uses to get information about users.


dtenenba:
LGTM but let's let @vortexing review it and merge if it looks ok....

PR #218, opened 2019-01-08
First pass at adding the flow cytometry instructions from the Core

vortexing:
This is the first pass of information from the Flow Core that I got in November 2018 but never got around to putting up.   I had to reformat several Word docs they had so this needs more formatting clean up and de-duplication, but others could help with that too.


ptvan:
Sure, I think that's reasonable, saves readers from extra clicks

PR #234, opened 2019-01-17
New information on parallel computing

atombaby:
Hi-

I'm opening this pull request to get some help and advice on where I'm going with this page/topic.  It won't be ready to merge for a little while, but I really wanted to get some opinions from y'all on what I've got, where I think I'm going, and if the level of detail is appropriate.

I've started with a generic introduction on parallel computing- primarily highlighting the difference between threading and distributed and will then do non-language specific discussions on specific approaches to making work run in parallel.  I'm planning on additional pages attached to this one that then implement examples of those different approaches.

I'd like to get a little more content before merging this in, so if you want to contribute, feel free to edit on this branch.

Thanks

### DIFF
--- a/_compdemos/cluster_parallel.md
+++ b/_compdemos/cluster_parallel.md
@@ -45,4 +45,34 @@ the threaded solution).  However, if there are many steps the resources on a
 single system will be a bottleneck, which makes a distributed solution more
 appealing.
 
+## An Atlas of Computational Workloads
 
+### Pleasantly Parallel
+
+"Pleasantly parallel" work (AKA "embarassingly parallel") is typically made up of many _completely_ independent steps.  By independent we mean that:
+
+  - any one step does not depend on the output or completion of any other step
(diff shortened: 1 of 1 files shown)


vortexing:
I think it'd be good to put a section on parallel computing in this page:
https://sciwiki.fredhutch.org/computing/cluster_rhinoGizmo/

But we should do that AFTER that pull request gets merged b/c I don't want to cross the streams unnecessarily.

vortexing:
@atombaby and @rbarfield should we merge this in?  I can then later go back and add an introduction to it in the clusteR_rhinoGizmo file, but I'm hesitant to update from the master at this point b/c I don't want to break anything.

atombaby:
Yes, I'd say let's merge, but I'd like to leave the branch as there's still lots of information to add.

rbarfield:
Looks good. Sorry for the late reply.

Issue #230, opened 2019-01-15
Tutorial on how to install R packages from GitHub


ptvan:
**Proposed Domain**
Bioinformatics

**Content Summary**
A tutorial on installing R packages from GitHub, with relevant code snippets

**Local Content Expert(s)**
Anybody really

PR #243, opened 2019-01-29
Update Bioinformatics R page to include GitHub installs

ptvan:
This addresses issue #230

### DIFF
--- a/_bioinformatics/prog_R.md
+++ b/_bioinformatics/prog_R.md
@@ -45,11 +45,14 @@ There are a number of available resources built on R that are free and open sour
 
 [R/Bioconductor events page](http://www.bioconductor.org/help/events/)
 
+### CRAN
+[CRAN](https://cran.r-project.org), (Comprehensive R Archive Network) is a public repository of numerous R packages along with R itself. Numerous staistical packages are available, though packages are not vetted as heavily as Bioconductor. 
+
+### GitHub
+[GitHub](https://github.com/trending/r) hosts many opensource R packages. As they are not necessarily vetted or peer-review, these packages can be more experimental than those on CRAN or Bioconductor. Instructions for installing R packages from GitHub are [here](http://kbroman.org/pkg_primer/pages/github.html)
+
 ### The Tidyverse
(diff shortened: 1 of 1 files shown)


dtenenba:
For instructions on installing an R package from GitHub, you link to an external page. Unfortunately that page mostly discusses other topics (how to *put* an R package into GitHub) so readers may be confused. Also there is no anchor to link directly to the relevant header. I would suggest maybe just pulling in the relevant text from that link rather than linking to it.

I also think the verbiage should be a bit stronger--we should emphasize that users should ONLY install from GitHub if the package (or a cutting edge version of it that's absolutely essential to have)  is not available in CRAN (or already available as part of an easybuild module that SciComp has put together). 
Otherwise what happens is people install weird unsupported versions and complain to SciComp and it may take SciComp a while to notice that they have installed a version that is ahead of what's on CRAN/BioC/Easybuild. 

Also, one silly trick is, you can't use the CRAN instructions to install a BioC package, but you CAN do the reverse. So if you have a bunch of packages to install, and some are in BioC and others are in CRAN, you can install them all in one command line if you use the BioC installation mechanism.

Finally, one more trick/recommendation that is not mentioned on the page you link to:

In the case where you want to install a cutting-edge version of package X (where the package is available in CRAN or Bioconductor but the latest version is only available in GitHub), I suggest installing the package using the standard CRAN/Bioconductor install instructions, and THEN installing the new version of it using GitHub. The reason is that in my experience, devtools::install_github() does not do a great job installing dependencies. So might be good to document that. 

So I'm not sure...maybe we want another separate page that tells you how to install packages from all these different sources? Or we could pull it into the page we are discussing now, but that might cause it to go into the weeds.

ptvan:
These are all excellent points @dtenenba. @rbarfield @vortexing how detailed do we want to get on this ?

vortexing:
I will take a stab at adding to this page.  I think this page could be used to explain just the types of issues @dtenenba brought up.  I think it's the perfect place.  Let's discuss how to behave locally, then how to behave on rhino/gizmo and then people will do better.  It'll take a while, but let's all start contributing edits on this and we'll get it done/review it as we go!!

vortexing:
@ptvan what do you think about this PR?  Have we done what we needed to do?

ptvan:
I think it looks good for the Bioinformatics page. I do want to preserve @dtenenba's tricks (installing CRAN/BioC first, then GH to get dependent packages, and the relationship CRAN <-> BioC), but maybe that's better on a separate "R tips and tricks" page ?

PR #253, opened 2019-02-07
remove costs

vortexing:
remove explicit references to storage costs from storage pages. Perhaps @dirkpetersen can link to a centernet page containing the current, maintained costs list so we can only manage that in one place?


dirkpetersen:
I am not sure why we want to remove them ? Has there been a discussion about it?

vortexing:
Cool!  So @dirkpetersen will go through the storage section and make sure every type of storage that has a cost to PI's has it's cost listed and correct. ;)

Issue #251, opened 2019-02-06
economy api Missing Links


rbarfield:
**Proposed Domain**
https://github.com/FredHutch/wiki/blob/master/_compdemos/EconomyCommand-API.md

**Content Summary**
There is section of the page that is missing a few links. Should potentially be added? Currently reads

*please check our HPC and data migration tutorials in the scicomp wiki at http://scicomp.fhcrc.org and the swc page on github for additional details:

- How to migrate large data to Economy File
- How to use Economy File and Scratch with Sequencing Data
- [swift commander on github](https://github.com/FredHutch/swift-commander)​
- [Check out this video how to the swc command](https://asciinema.org/a/17172)​
- HPC: using Economy File and Scratch on Gizmo*

Are the three that missing links pages that need to be developed or are they just missing? 

**Local Content Expert(s)**
@vortexing @FredHutch/scicomp

dirkpetersen:
Fixed a link:  https://github.com/FredHutch/wiki/pull/255

atombaby:
These links look to be fixed

PR #255, opened 2019-02-08
a couple of storage fixes.

dirkpetersen:
moved doc from old wiki 
how to migrate large data to Economy Local

and a few fixes

### DIFF
--- a/_compdemos/EconomyCommand-API.md
+++ b/_compdemos/EconomyCommand-API.md
@@ -111,11 +111,9 @@ now let's download a subfolder of that folder /test to our scratch file system f
 please check our HPC and data migration tutorials in the scicomp wiki at http://scicomp.fhcrc.org and the swc page on github for additional details:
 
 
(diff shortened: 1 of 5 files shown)


_compdemos/EconomyCommand-API.md
```diff
@@ -307,11 +308,11 @@ instead of using the swift command from the unix shell one can also access swift
 
 
 
-​## access from the command line or Rest API
+## access from the command line or Rest API
```
atombaby: This section is rather unclear to me- it seems that there may be some missing information about what those access methods are or what "direct access" is.  I think we also need a link for the reference to"How to Access the Swiftstack Object Store"

vortexing:
Relates to issue #251.

vortexing:
@rbarfield do you want to take a read of this to add any comments or edits?  Is it starting to address your issue?  I think work is ongoing but before we merge this we should all weigh in.

PR #256, opened 2019-02-11
Working with S3 and Scratch

vortexing:
I forgot I had this little demo.  I put it in bioinformatics demo, but if y'all SciComp guys want to adopt it and make it yours, feel free!


dtenenba:
Nice. I suggest saying "Using Economy Cloud storage" instead of saying "Using Economy storage", as Economy (local) storage is something different (swift).

PR #264, opened 2019-02-21
Issue 239 - env modules.

bmcgough:


vortexing:
@bmcgough Let me know what you think about my reorg.  NoMachine content was duplicated in Access Methods, and technology stuff should go over on the resource_overview markdown so that this page can focus on scientific software.

PR #276, opened 2019-03-12
Update EconomyCommand-API.md

dirkpetersen:
updated info on how to access swift from R and aws cli


vortexing:
When I have another moment I'll come back and make some organization suggestions/de-deuplication work.  Perhaps @dtenenba also might have some insight as to how to document all of the access methods to S3 and Swift since some are the same.  Maybe we need a slightly different organization of demos (such as possibly a boto3 demo that explains how to get to Swift or S3 using it, then an R/aws.s3 demo that has both, then a CLI, then a client, or something to that effect).  We just need this to be more clear and less duplicated/jumpy.

vortexing:
I'm working on your docs here @dirkpetersen.  YOU even have docs saying how to make a container in Swift?!?!?!?!?!!

vortexing:
LEt's get this in so that the stuff is correct and more docs later.

Issue #282, opened 2019-03-18
Documentation on Nextflow


sminot:
**Proposed Domain**
Computing

**Content Summary**
How to use Nextflow on AWS with configuration details specific to Fred Hutch

**Local Content Expert(s)**
Sam Minot can contribute the bulk of this content

vortexing:
Done!

PR #283, opened 2019-03-18
Add details on Nextflow

sminot:
This is a start to some of the details on how to use Nextflow #282


vortexing:
As is often the case, I say, BOTH!!!  I am going to do a bit of Nextflow work today so I had intended to have this open, go through it, add edits, and possibly move anything more detailed than "What is Nextflow and why should I use it" into a detailed demo file.  That cool with y'all!?!

vortexing:
This will be closed by the merge of PR #316

PR #288, opened 2019-03-28
add aws creds test

vortexing:


iankirkman:
I have aws credentials now, but I do not have a pi bucket and am denied access from creating an s3 bucket in cli. So half of the test works for me! 

I'm also going to commit a minor change to the ssh instructions because windows ssh actually requires the same username format as mac.

PR #309, opened 2019-04-26
first pass compute merge

iankirkman:


vortexing:
I made some tweaks and I think this is reasonable to get moved into the site in it's current form.  However github now things I've contributed too much too and thus it's not wanting me to review it.  Perhaps a final review by @atombaby or @bmcgough  or @dirkpetersen  or @dtenenba  might be good?  Just a skim?

atombaby:
So the new branch is looking good.  The `_bioinfcomputing` directory is replacing `computing`, right?  Is there a reason we're leaving that around?  Or have I misunderstood that?

vortexing:
Yup, @iankirkman is using the bioinformatics content as your introductory material essentially.  The tension between the two domains is finally done!

PR #316, opened 2019-05-03
cloud computing merge with new nextflow resource page

iankirkman:


sminot:
I added a couple of details to reflect the Nextflow module that John Dey is currently building. Notably, I wanted to remove the reference to users building their own version of Nextflow, as I do think that the working version will be available shortly.

sminot:
@vortexing @iankirkman I appreciate your work in getting these details added to the wiki before Tuesday. Thank you!

vortexing:
Fixed some links and I think we can put this in!!!

PR #329, opened 2019-05-26
Updated Tensorflow docs

dirkpetersen:
Fixing some GPU/Tensorflow docs since tensorflow-gpu 1.13.1 can now install from pypi and GPUs on Gizmo J nodes are active


atombaby:
I am very reluctant to integrate this documentation until we get the GPUs configured by Chef and integrated into Slurm.  This will change the interface and methods used to select and run GPUs.

atombaby:
I've updated the Slurm commands to include the flags necessary to request GPUs from the scheduler.

I'm a little concerned about the number of changes in the commit- I may have merged changes from master backwards.  The diff between this and master looks correct though.

vortexing:
I do think Ian must have updated from master at some point but that makes the review process a little challenging but Github will merge it a-ok.

PR #333, opened 2019-05-30
adding scicomp contact info

dirkpetersen:
just quick change which would allow me to put sciwiki into my email footer and not any other info about scicomp support 

I'd like to put this 

https://sciwiki.fredhutch.org/scicomputing/

instead of this 

https://sciwiki.fredhutch.org/scicomputing/comp_index/

into my footer but currently it does not redirect and gets a 404


vortexing:
I believe this was resolved with a hack.  Dirk, reopen if it was not.

vortexing:
oh but you did also make changes.

PR #344, opened 2019-06-13
Changes to GPU documentation

atombaby:
I moved @bmcgough 's table from the overview to the GPU-specific documentation and added a column to the overview table for the GPU data.


atombaby:
@bmcgough can you check the table formatting, make sure I've got that right?  Thanks

PR #361, opened 2019-07-18
compute platform refinements

vortexing:
So this page: https://sciwiki.fredhutch.org/scicomputing/compute_platforms/

Needs help.  The autogenerated tables are great for maintaining updates in one place, BUT they do not allow for any explanatory text in between the tables.  Perhaps instead of looping through all the parts of the yml and making multiple tables at a time with no text in between, could we break that up a bit so we can explain what abbreviations and such are IN the table?  Also, why that table content is informative or interesting and how to use it?  

@bmcgough This is your baby so I am hesitant to fiddle with it at all.  Plus I don't actually know what some of the contents of the tables even are. ;)

### DIFF
--- a/_scicomputing/compute_platforms.md
+++ b/_scicomputing/compute_platforms.md
@@ -1,29 +1,33 @@
 ---
 title: Supported Technologies and Platforms
-last_modified_at: 2019-04-26
+last_modified_at: 2019-07-18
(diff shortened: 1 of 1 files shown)


vortexing:
@bmcgough

vortexing:
Fingers crossed!

PR #366, opened 2019-08-01
Add documentation on microbiome analysis tools

sminot:
I'd like to add a page describing some tools available for microbiome research. I'm not sure how to get this new page to show up in the right place -- I hope this is correct!

### DIFF
--- a/_compdemos/microbiome_tools.md
+++ b/_compdemos/microbiome_tools.md
@@ -1,0 +1,97 @@
+---
+title: Bioinformatics Tools for Microbiome Analysis
+last_modified_at: 2019-07-31
+primary_reviewers: sminot
+---
+
+Data generated from microbiome experiments tends to require a distinct analytical
+approach, one that takes into account the many different organisms which may be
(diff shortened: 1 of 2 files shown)


sminot:
I did put a small mention in the Nextflow page -- were you thinking something larger or is that enough @vortexing ? Is there another main page I can use to add a link to this? I was also thinking about linking from the other microbiome pages on other parts of the Fred Hutch website

sminot:
Oh, also I think that my other PR has to be merged before this one can be, since I based this one off that one.

vortexing:
hmmm.  I did merge the other one and now it says there are conflicts here.  

As for links I'm not really seeing where in the main articles we could really link to it now that we've taken out the Bioinformatics subdomain.  That domain was really where I had hoped to put stuff like this. hmmm.  I saw that you have a link in NExtflow and also it's searchable so maybe it's not that big a deal.  Just keep putting links out there in the world so the right people can find it!  Let me see if I can resolve these conflicts.

sminot:
Thank you @vortexing !

PR #368, opened 2019-08-01
Adding quick start guide for large scale computing

k8hertweck:
Addresses #353 . Addition of this new page to sidebar headings will be included in #348.

Issue #348, opened 2019-06-25
Software Development overview 


k8hertweck:
**URL of page(s) of concern**
https://sciwiki.fredhutch.org/scicomputing/software_overview/

**Describe problem**
Sidebar shortcuts not aligned with subheadings

k8hertweck:
Issue resolved by #393

PR #370, opened 2019-08-01
Updating navigation and descriptions on overviews

k8hertweck:
Addresses #348 by cross checking the sidebar subheadings across current page content. Includes Quick Start Guide from #368 so subheadings will still be complete when that's merged.

Additional changes: Includes a few minor updates to some subheadings to improve readability. Also removes the subheadings from the Data Generation and Scientific Computing overviews (so we don't need to update things in two places). Revised the summaries on the latter, given the subheadings are no longer present.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -104,10 +104,10 @@ scicomputing:
     children:
       - title: "Overview"
         url: /scicomputing/access_overview/
-      - title: "Methods"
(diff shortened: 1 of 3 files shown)


k8hertweck:
I'm happy to replace the subheadings; it might also be useful to include mention of subheadings in each overview page which don't link out to another wiki page (either content that is only on overview, or to an external site, like another repo).

PR #371, opened 2019-08-02
update snail tidbits

vortexing:
Updating tidbits about using snail.fhcrc.org.  Note:  Before we merge this pull request, we should figure out why the instructions for adding two lines to config so you don't have a two step process doesn't work/needs more detail to make it work.


vortexing:
@FredHutch/scicomp any final input on Snail?

PR #378, opened 2019-08-22
adding matlab to common languages

k8hertweck:
addresses #347


k8hertweck:
We've got approval from #347, so this is good to merge

PR #418, opened 2019-10-21
Add Nextflow group to list of Community Groups

CowardlyBattleCat:


k8hertweck:
Thanks @CowardlyBattleCat ! 

@Chilliwack we should also add Nextflow to https://github.com/fredhutch/community_groups

Issue #365, opened 2019-07-31
Largenode tidbits


vortexing:
**Proposed Domain**
https://sciwiki.fredhutch.org/scicomputing/compute_jobs/

**Content Summary**
Discuss more about how to use largenode.  Looks like you HAVE to ask for memory, and the minimum memory you need to ask for to get it to go is 42+GB?  Perhaps add a largenode example?  Also, maybe an example in a bash script header rather than a one liner?  
```
#!/bin/bash
#SBATCH --output=Logs/%x_%A_%a.out
#SBATCH -p largenode
#SBATCH -c 6
#SBATCH --mem=43G
```  

**Local Content Expert(s)**
SciComp crew

bdconnolly:
@vortexing I have added a example about submitting a job to largenode partition to https://sciwiki.fredhutch.org/scicomputing/compute_jobs/#examples

Is this what you were looking for?

bdconnolly:
This issue has been fixed in merge request #422

PR #422, opened 2019-10-24
Fix formatting and largenode example

bdconnolly:
- Issue #365 
-- In Examples section put new headers to describe each example. That way "Use largenode partition" shows up in the right hand pane of links. 
- Formatting changes: Examples header was H4 instead of H3

### DIFF
--- a/_scicomputing/compute_jobs.md
+++ b/_scicomputing/compute_jobs.md
@@ -164,7 +164,7 @@ a) there are no idle resources for your job,
 b) jobs running under your account are already consuming the maximum amount of resources available for accounts, and 
 c) your job does not have a high enough priority to be run.
 
-If your job won't run because resources aren't available, the reason code will be "Resources".  Jobs in this state have resources reserved for them (a "priority reservation") and should run next when those resources become available.  This reason is only applied to one or two jobs in the queue.  Every other job will have the reason "Priority," which indicates that the job does not have sufficient priority to have resources reserved (reason "c" above).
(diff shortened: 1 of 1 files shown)


_scicomputing/compute_jobs.md
```diff
@@ -147,17 +149,22 @@ requested time.
...
+The three most common reasons are:
+a) there are no idle resources for your job, 
+b) jobs running under your account are already consuming the maximum amount of resources available for accounts, and 
+c) your job does not have a high enough priority to be run.
 
-If your job won't run because resources aren't available, the reason code will be "Resources".  Jobs in this state have resources reserved for them (a "priority reservation") and should run next.  This reason is only applied to one or two jobs in the queue.  Every other job will have the reason "Priority," which indicates that the job does not have sufficient priority to have resources reserved (reason "c" above).
+If your job won't run because resources aren't available, the reason code will be "Resources".  Jobs in this state have resources reserved for them (a "priority reservation") and should run next when those resources become available.  This reason is only applied to one or two jobs in the queue.  Every other job will have the reason "Priority," which indicates that the job does not have sufficient priority to have resources reserved (reason "c" above).
```
vortexing: "This reason is only applied to one or two jobs in the queue."
Is this what you truly mean?  Or do you mean something more like, "Typically, only a small number of jobs should be in the "Resources" state if the jobs are submitted with reasonable resource requests.  A large number of jobs in this state may indicate something is a miss with your request..." Or something like that.

PR #442, opened 2019-11-08
Fix link for howtoRhino page

CowardlyBattleCat:
`howtoRhino.md` and `howtoGizmo.md` are both in `_compdemos/` but the Rhino link pointed to the `_scicomputing/` dir instead. I fixed the link, but it's possible that both files are in the wrong directory instead.


vortexing:
How to Gizmo likely needs to be converged with the `Job Management` article now. 
 https://sciwiki.fredhutch.org/scicomputing/compute_jobs/

How to Rhino likely should stay and is essentially a portion of the "quick start" guide any lab might make for their own lab members for which the onramp to Rhino is important (which is not everyone).  

So link fix is in this PR which should get merged, but I will file a separate issue to address the possible convergence above.  ;)

PR #446, opened 2019-11-12
We need to simplify. Streamlining will be easier that way.

vortexing:
There apparently are many ways to skin a cat, I just don't know why people know that that are.  As I was reviewing issues, much of the streamline/simplify requests are actually what the original overview pages were meant to be.  I think in an effort to continue to improve things, streamline them for readers AND reduce duplicated text (since we need this to be easier to maintain), some of my consolidations are proposed here.  

I have more suggestions now that content is getting more mature but I think it's best to bite off smaller pieces at a time.  We are all busy after all.


vortexing:
This has become too complicated and I think it will break stuff.  We'll get these changes in in smaller chunks in teh future.

PR #450, opened 2019-11-14
Update matlab docs

atombaby:
Recent conversations at office hours highlighted problems with the matlab docs.  I've updated these docs based on those interactions and problems we'd found following them.


vortexing:
I added a few little tweaks.  

My only question here is that at this moment, when I read this, I wondered, what IS a "Linux session manager"?  I wonder if the intended audience would know?

@fizwit if this is cool by you, review it and let's get in the updates!

PR #451, opened 2019-11-14
Adding content to code examples and templates (Software Development sections)

k8hertweck:


vortexing:
@k8hertweck should this get merged? Is it ready?  I'm likely not the right person to review.

k8hertweck:
I mean, the content isn't ground breaking or anything, it's mostly organizing things. The only possible issues would be:
- broken links, or links to private repos
- recommending repos from the Fred Hutch org that aren't useful (I only selected things that have some documentation and/or are in current development)

I just checked the links and it seems like things are accessible. Could you take a quick look at the last section on Fred Hutch code examples, and if it doesn't seem like crap, then I'm happy to merge?

PR #454, opened 2019-11-19
refresh shiny!!!

vortexing:
@dtenenba I just re-read the instructions as I was prepping to request deployment so I thought I'd tune up how it reads. 

I didn't change any substance, only formatting, a little order, and clarifying.

### DIFF
--- a/_compdemos/shiny.md
+++ b/_compdemos/shiny.md
@@ -53,7 +53,7 @@ There are two ways to set up your repository and either is suitable.
 
 ### From a New GitHub Repo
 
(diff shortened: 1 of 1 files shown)


dtenenba:
I like it! Great how you spell out a lot of things that are not necessarily obvious. Telling people to test locally should save us some bug-hunting.

The link to github.com is not fully qualified (needs `https://` in front of `www.github.com`, and probably doesn't need the `www`).

vortexing:
Great! Fixed it!

PR #465, opened 2019-12-10
deleted Software Dev group stuff and renamed Py Group

Chilliwack:
deleted the info about the software development group as it's going away and renamed the python coding group to Python User Group


k8hertweck:
Can we go ahead and update the other relevant groups to "R User," "Nextflow User," etc too while we're at it?

Chilliwack:
updated the other groups as well in 2c9480c

PR #473, opened 2020-01-07
Nextflow FAQ

eharkins:
@sminot and I have added some "gotchas", FAQ, and more info to the nextflow docs. Hopefully we can set up a system where folks fork and PR changes to this doc when they have generally applicable nextflow solutions that others can benefit from here at the Hutch. That way people aren't running into the same issues as much or doing as much re-asking of questions in the nextflow channel on slack.

### DIFF
--- a/_compdemos/nextflow.md
+++ b/_compdemos/nextflow.md
@@ -4,7 +4,10 @@ last_modified_at: 2019-10-28
 primary_reviewers: sminot
 ---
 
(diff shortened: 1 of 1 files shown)


dtenenba:
This looks really good. Wondering if it would be good to mention [Tower](https://tower.nf).
Guessing most potential Nextflow users would appreciate the UI showing their jobs' status.

eharkins:
Good idea @dtenenba! I haven't used it, would you like to add something about if you have?

sminot:
I have used Tower and I can add some text. Great idea @dtenenba!

sminot:
Ok, just added some text describing Tower, @dtenenba

dtenenba:
@sminot be sure and mention how to actually invoke tower. Don't you have to submit your job with a `-with-tower` flag or something?

sminot:
That's actually taken care of with the additions to the config file, as they describe it on the Tower website

dtenenba:
OK, nice!  So that probably turns on tower logging for all jobs. I can imagine you might want to turn it off sometimes so it might be good to discuss how to do that. But that's not a big deal, so I will go ahead and merge this.

PR #491, opened 2020-02-20
Adding more stubs for the two workflow managers

vortexing:
I thought it'd be nice to put in some hooks to eventually elaborate on about workflow managers where they touch the existing content.


sminot:
Do we want to point people towards using SLURM + workflow managers? I haven't tried that for myself here, and so I'll defer to you, @vortexing. Not sure how you feel about Singularity, or easyload, or if there is a general best-practice way to approach SLURM + workflow managers.

vortexing:
Yup, we very much do.  Cromwell can make SLURM sing and that's what a lot of people might use even if they don't need/want to go to the cloud.

vortexing:
Also, all my config work is specifically to use easybuild modules on gizmo as if they were docker containers. ;). Super smooth!  However since there are no Nextflow + SLURM docs, maybe I should just link to Cromwell?

sminot:
Oh great! Well if people can come across this and realize that they can use Cromwell to make SLURM sing, that sounds awesome to me!

Issue #502, opened 2020-03-11
Repurposing Python User Group's: Python IDEs content for SciWiki


Chilliwack:
**Proposed Domain**
Specify where should the new content should go (Data Generation or Scientific Computing), including a recommended subsection.

Scientific Computing > Software Development > Python

**Content Summary**
Summarize the topic you'd like to be added, or include a URL with content you would like to be expanded. 

Repurpose and showcase the content created by KB, ZR and JB listed here:

VSCode: https://github.com/kmayerb/visual_studio_code_demo/blob/master/README.md
PyCharm: https://github.com/zyd14/pycharm_demo/blob/master/README.md
Jupyter Lab: https://github.com/Chilliwack/jupyter_demo/blob/master/README.md

And focus on the fact that as a new user you'll want to get familiar with a debugger, variable inspector, linter, version control integration, text editor, and basic keyboard shortcuts regardless of the IDE to be effective. Open to suggestions as well on how to repurpose this content for the wiki.

**Local Content Expert(s)**
Suggest any Fred Hutch based experts who we might ask to contribute (GitHub ID is preferred, but name of someone and/or desired expertise is ok too).
@zyd14 @kmayerb @Chilliwack

vortexing:
I would seriously consider just straight up adding each of these three READMEs into wiki/_compdemos/ to render in the Resource library.  Then add links to each of them form the Scientific Computing > Software Development > Python page.  I can make that happen with the correct front matter if it is unclear how to do this. ;). Just tag me.

Chilliwack:
great idea but these are my meal ticket into the docathon so if no biggie Ill wait till then to do it :)

vortexing:
Ah, well a better plan would be better.  I was going minimalist/low bar...

vortexing:
NUTS!  The pycharm images are still broken on the live site.  https://sciwiki.fredhutch.org/scicomputing/software_python/#pycharm

@zyd14

PR #519, opened 2020-04-08
Python IDEs

vortexing:
I took a first pass at copy/paste to fix up the Python page.  I think it'd be great to get input from @kmayerb and @zyd14 initially to begin to better organize and curate the material to make it flow a bit better, maybe that means make it more article than demo (I'm assuming your original repo's will still exist, this is just an article-form version).  Once that is done, if you guys wouldn't mind deciding on how to handle your images that are remaining in the text into the right assets location in the Wiki repo (assets/python/) or good links to yours?  

Then we can open it up to a real PR instead of a draft and get some other people's thoughts on it.  Sound ok?


vortexing:
Closes #502

vortexing:
So what do you guys think?  Should we merge this in and see how it looks all rendered and decide if we should be breaking it up into IDE Resource Library entries for each IDE and then leave the intro material in this article or not?  @zyd14 @kmayerb   I will defer to you guys on this.

vortexing:
I guess I mean too that I think that sort of structure is likely what @zyd14  was going for in this other branch: https://github.com/FredHutch/wiki/blob/enhance/software_python_szr/_scicomputing/software_python.md

Except instead of linking out to your repos for ALL the details, I'd like at least most of the details to live in the Wiki somewhere (either in this page OR in separate pages in _compdemos), so that if you delete your repo, the content still exists here).

zyd14:
@vortexing - I think seeing it rendered together in one page would be nice... if it ends up looking like too much for a single page we could break it up as you advised.  It'd be nice to avoid making people jump around too much if it's not necessary.

vortexing:
Sounds good.  It's always harder to get the content CREATED and not duplicated FIRST, and then if we want to parse it out a bit, that's fine.  I mean ,people can scroll and there's a TOC, so rather than making them hunt, this seems like a good starting point.

vortexing:
I won't delete the branch yet until we know for sure what we are doing.  It should be live now (or in a few minutes).

zyd14:
Not sure why the images are being displayed when you view it on github but not on the webpage... I'll look into it and fix it up

PR #526, opened 2020-04-18
added hardware specs

dirkpetersen:


dtenenba:
I think it should be "an" before NVIDIA (I pronounce it 'enn-vidia' in my head).

PR #527, opened 2020-04-19
gizmo checkpointing first try

dirkpetersen:
very first beta attempt checkpointing


vortexing:
@dirkpetersen  or others, where would you imagine this content to be linked to in the main article section?  Would a brief mention "We discuss checkpointing here." be reasonable to put in, say the compute_jobs page? 

https://sciwiki.fredhutch.org/scicomputing/compute_jobs/
It'd be best to make sure that before we add a new demo page, that there is some mention of it in the main site so people can (better... hopefully) find it.

dirkpetersen:
It has not been linked yet but is now ready for wider testing

PR #543, opened 2020-05-07
Singularity updates

scottsisco:
This is a first draft

### DIFF
--- a/_compdemos/Singularity.md
+++ b/_compdemos/Singularity.md
@@ -5,60 +5,107 @@ main_author: Dirk Petersen
 primary_reviewers: bmcgough, dirkpetersen
 ---
 
+## Why use Singularity
+
 Docker can only run containers as root so we cannot use them in shared multi-user environments such as `Rhino`/`Gizmo` with shared storage such as `/home`, `/fh/fast` or `/fh/scratch`. Singularity can import Docker containers and run them. 
-
-## Understanding Singularity
-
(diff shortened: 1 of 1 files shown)


_compdemos/Singularity.md
```diff
@@ -5,60 +5,107 @@ main_author: Dirk Petersen
...
+```
+Import the Docker container and build the Singularity container.
+```
+singularity build pickanameforyourcontainer.sif docker://point-to-dockerhub-repo
 ```
-    petersen@gizmok13:~$ ml Singularity
-    petersen@rhinok13:~$ singularity exec ~/ubuntu-python.sif python3 /fh/fast/.../script.py
+Run the container.
+```
+singularity exec pickanameforyourcontainer.sif container-command
```
atombaby, line 33: It's not _required_ that one build the container before running it (AFAIK).  One good thing about separating `build` from `exec` is that the long-ish build phase is seen as distinct.  It may be confusing if (for some reason) the cached build is unavailable and singularity rebuilds it.

scottsisco: @atombaby The build step is required if converting a docker container to a singularity container. The reason I went with the build command then exec command is Dirk thinks and I agree with him that the main way people will use singularity is to convert a docker container to a singularity container. 

Maybe I should add a line that says build is only necessary if you have not already created a singularity container from a docker image or if you are looking to update your image to the newest docker container?

PR #560, opened 2020-05-15
remove references to rstudio via X11 or rstudio.fhcrc.org

dtenenba:
Preferred way to use rstudio is via launch_rstudio_server script.

### DIFF
--- a/_scicomputing/software_R.md
+++ b/_scicomputing/software_R.md
@@ -34,24 +34,58 @@ When using R/RStudio locally, you have the option to install a number of differe
 
 
 ## Remote (`Rhino` and `Gizmo`) Use
-If computing resources beyond what is available via your desktop are required, you may consider running R scripts or RStudio from the `rhinos` or `gizmo`.  When using R/RStudio on shared computing resources, different options for builds and modules are available that you can take advantage of.  `SciComp` makes pre-built modules available for researcher use in order to facilitate more reproducible and reliable use of software on the local cluster.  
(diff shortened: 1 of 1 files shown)


_scicomputing/software_R.md
```diff
@@ -62,17 +88,11 @@ You can run this with any version of `R` that is available on our shared computi
...
+* Once you are on the node you grabbed, choose a version of `R` by using the [module load](/scicomputing/compute_environments/#how-to-use-environment-modules) command (`ml` for short). (Example: `ml R/3.6.1-foss-2016b`).
+* Run the `launch_rstudio_server` command. This will produce a URL that you can paste into your browser. (This URL only works inside the Hutch network, so you need to be on campus or using VPN.)
 * In your browser, log into RStudio using your HutchNet ID and password.
-* When you are finished using RStudio Server, you can
-terminate it by typing this command on the node you "grabbed" (the same machine where you launched RStudio Server): `launch_rstudio_server --kill`
+* When you are finished using RStudio Server, you can terminate it by typing this command on the node you "grabbed" (the same machine where you launched RStudio Server): `launch_rstudio_server --kill`
 * Alternatively, you can just wait for your `grabnode` allocation (the number of days you specified when grabbing the node) to expire, and RStudio Server will become unavailable after that.
```
atombaby, line 96: Kind of feel just letting sessions idle out is wasteful and not a good alternative.  It is likely important to note that the rstudio server will be unavailable when grabnode kills it.

_scicomputing/software_R.md
```diff
@@ -34,24 +34,50 @@ When using R/RStudio locally, you have the option to install a number of differe
...
+module spider R
+```
+
+Choose one and invoke it with `ml`.  For example:
+
+```
+ml R/3.6.2-foss-2016b-fh1
+```
+
+Note that you can use tab-completion with the above command. For example, you can type
```
atombaby, line 56: Eventually might be nice to pull out `module` and tab-completion out into its own page.  Having it here seems a bit of a distraction to the main point of the page.

Don't need to fix in this PR tho...

PR #575, opened 2020-06-01
document uploading files to s3 for use with UCSC Genome Browser

dtenenba:
There have been several requests by people who want to use the UCSC Genome Browser with their own data and it would be nice to be able to point people towards documentation that already exists, so here's an attempt at it.

Undoubtedly it needs some work but the bare minimum of information is there.

I'm also not sure where it should go and defer to @k8hertweck and @vortexing. @atombaby suggested compdemos so that's where I put it, but let me know if it should move.

In the meantime, I actually have someone who needs this and and I will just point them towards the page in this PR until it is merged.

### DIFF
--- a/_compdemos/ucsc-track-s3.md
+++ b/_compdemos/ucsc-track-s3.md
@@ -1,0 +1,81 @@
+---
+title: Uploading Tracks to view with the UCSC Genome Browser
+main_author: Dan Tenenbaum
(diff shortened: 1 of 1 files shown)


k8hertweck:
I think compdemos is a great place for this. I'm going to think about other places it could be linked in the main wiki, but I'm happy to merge since it looks good!

PR #585, opened 2020-06-10
Maxquant on linux docs

dtenenba:
The MaxQuant proteomics software is now available on linux, so the cloud- and Windows-based pipeline we had before is now deprecated in favor of running MaxQuant on linux on the gizmo cluster, where it will not incur any costs.

This is documentation intended for users of that previous pipeline (or really anyone who wants to run MaxQuant).

### DIFF
--- a/_compdemos/maxquant.md
+++ b/_compdemos/maxquant.md
@@ -1,7 +1,7 @@
 ---
 title: Proteomics using Maxquant on Gizmo
 main_author: Dan Tenenbaum
-primary_reviewers: 
+primary_reviewers: bmcgough, fizwit, atombaby
 ---
 
 This demo gives instructions for running the MaxQuant proteomics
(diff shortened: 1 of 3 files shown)


_compdemos/maxquant.md
```diff
@@ -0,0 +1,234 @@
...
+
+```xml
+   <numThreads>24</numThreads>
+```
+
+Remember the number you changed it to; you will need it later.
+
+### Submit job to cluster
+
+The rest of these steps need to be done on a Linux HPC system. Using `ssh`, `PuTTY`, or `NoMachine`, connect to the system `rhino01`.  For more information about these connection methods, read [Computing Resource Access Methods](/scicomputing/access_methods/).
```
atombaby, line 98: The "rhino" alias can be used for ssh and putty, nomachine needs the explicit name.  Maybe punt and just remove ssh and putty statement leaving the link to the access methods page?

_compdemos/maxquant.md
```diff
@@ -0,0 +1,234 @@
...
+
+<!-- 
+TODO: after Bionic (Ubuntu 18.04)
```
atombaby, line 101: Might as well make that change now.  There's just a couple days left and as this is a new workflow it should use the bionic-flavor resources.

_compdemos/maxquant.md
dtenenba: Thanks. I still need to put in a machine name to connect to, somewhere. And I don't want to use "rhino" until that points to bionic machines. It's possible that people will use these instructions before the changeover on Tuesday.

PR #586, opened 2020-06-11
zsh instructions for modules

matsen:
I don't use modules extensively, but this has worked for me in the past, and so I thought I would add it here.


matsen:
Oh, I hadn't thought about that. I haven't tried it yet on the Bionic nodes but closing for now. I'll re-open if it doesn't work out of the box.

PR #592, opened 2020-06-18
small note about screen use on the rhino cluster

WhiteAu:
note about how screen interacts with rhino and method to more easily keep track of _screens_.


dtenenba:
...or just make a note of which rhino you are on, so you can get back to your session.

PR #594, opened 2020-06-18
Remove redundant pages from compdemos

atombaby:
howtoRhino and howtoGizmo (in the compdemos section) have mostly been made redundant by improved information in the scicomputing section.  howtogizmo has been removed, but howtorhino has additional information that should be moved into other pages


vortexing:
In the info that comes up when you log onto the rhinos, the HowToRhino page is one of the sciwiki links: 
```* To use this system most effectively, please read these guides:
    https://sciwiki.fredhutch.org/scicomputing/compute_quickstart/
    https://sciwiki.fredhutch.org/compdemos/howtoRhino/#rhino-compute-nodes```

vortexing:
@atombaby Perhaps we should consider what has been done in this branch vs what exists now.  Then if that can be merged, perhaps it will be easier to evaluate what other Bionic type changes we need to still do?  Thoughts?

atombaby:
Merged with #623

PR #600, opened 2020-07-09
Fixed local wiki build.

jefftucker:
Dependency conflict between html-proofer and nokogiri resolved by updating html-proofer to highest version with common nokogiri version dependency.  Also fixed issue with Faraday incompatibility by pinning faraday based on [this](https://github.com/github/personal-website/issues/166). 

Updated README with some extra information on how to resolve various local build errors that I encountered when building on a clean system under WSL.

Would be helpful if someone else could try building this locally (ideally under Max and Linux) and verify that it works.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :development, :test do
   gem 'jekyll-gist', '~> 1.5.0'
   gem 'jekyll-remote-theme', '~> 0.3.0'
   gem 'jemoji', '~> 0.10.0'
-  gem 'html-proofer', '~> 3.9.0'
+  gem 'html-proofer', '~> 3.14.0'
   gem 'rake', '~> 12.3.0'
+  gem 'faraday', '~> 0.17'
 end
--- a/README.md
+++ b/README.md
@@ -162,6 +162,8 @@ it can be found [here](https://www.ruby-lang.org/en/downloads/).
    
 1. You may need to install gems used by the site.
    Type `gem install -g Gemfile` to install all of the gems the site uses.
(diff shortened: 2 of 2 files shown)


dtenenba:
The instructions you give are specific to debian-like systems and won't apply to Mac, windows (non-WSL) or Linux (other distributions) users.

k8hertweck:
@dtenenba so the note in the README should specify this is a fix for only those systems? Do we know if local builds are an issue for other systems?

dtenenba:
Anyone might conceivably want to make a local build of the site just to test out things or see how they look.
I'm wondering if we should create a Dockerfile and associated docker image for building the site. Then you could build it on any OS that has Docker installed.....

I've had trouble building it before on Mac so I have done just that, but I never saved/kept/committed the Dockerfile I used.

dtenenba:
So I am having trouble with this, both with my mac and in the ruby:2.7.1 docker image:

```
root@953cf1a1803d:/work# gem install -g Gemfile
ERROR:  While executing gem ... (Gem::UnsatisfiableDependencyError)
    Unable to resolve dependency: user requested 'html-proofer (~> 3.9.0, ~> 3.14.0)'
```

k8hertweck:
I just succeeded in building locally (MacBook Air) with no modifications from the existing instructions.

A Dockerfile would be a great option (I'd use it for training), but in the meantime, I'm happy to keep documenting potential pitfalls as best we can.

k8hertweck:
@jefftucker was the update to `html-proofer` required for you, or were the missing packages the biggest hiccup?

Issue #605, opened 2020-07-20
"Invalid account" error


k8hertweck:
I just answered a help request for “Invalid account or account/partition combination specified” when using `grabnode`. This same questions has also been asked at least every six months in Slack. It seems reasonable to include this as a note in the wiki that folks need to email SciComp to get set up to use Gizmo.

PR #606, opened 2020-07-20
Add snippets about invalid account errors for new users

atombaby:
A common problem for new users is that a missing association to a PI account raises an error which isn't terribly clear about the reason or solution.  This adds additional information to key pages for new users.


k8hertweck:
Awesome, thanks for getting to this so quickly.

k8hertweck:
closes #605

PR #608, opened 2020-07-23
some updates on glacier features

dirkpetersen:
for Economy Cloud


vortexing:
So while I'm pro-disclosure, I wonder if adding these before it's "ready" is really beneficial or if that means that in the future someone will have to remember to come take it out. Should these hints be included in the SciComp announce section or the cloud re-architecture work that @jefftucker  is doing??  @k8hertweck

vortexing:
This also has no information about "Glacier features" so maybe this isn't done?

PR #623, opened 2020-09-17
Wiki updates for bionic upgrade

atombaby:
Now with beagle updates, 'cause beagle didn't get upgraded, just removed.


vortexing (approved):
This looks like a ton of work, so LGTM and then I will do a re-review once it's merged in to line up what issues can be closed and then find any teeny bits we've missed.

PR #625, opened 2020-09-17
koshu and beagle hunting

vortexing:
I found some beagles where they shouldn't be AND I archived some Koshu stuff until it can be better supported in the future times when things aren't insane... whenever that is.


atombaby (approved):

vortexing:
Not sure now why this still has commits that aren't in the other PR, BUUUUT I'll merge it and then we'll see what we have.

PR #631, opened 2020-09-18
Remove note about hybrid economy local

scottsisco:
I had someone ask about when the hybrid solution between economy local and fast will be available and sited this note. I have been told we are not going to be offering a hybrid model by Dirk so this should be removed.


vortexing (approved):
Dreamy.

PR #633, opened 2020-09-24
updating singularity command

scottsisco:
The three dots ` ... `at the end of the `singularity exec --nv docker://tensorflow/tensorflow:latest-gpu-py3 python3` command are making the command none functional for end users.


dtenenba (approved):
LGTM

PR #638, opened 2020-10-07
SSH key clarifications

k8hertweck:
Adding additional language to help users new to the cluster AND to computing.


vortexing (approved):
Let's do this.

PR #639, opened 2020-10-08
remove CWL

ilkerg:
CWL is not a workflow manager, it's just a language that some tools make use of.


vortexing:
On an anti-CWL mission?  But you're right, that's true.

ilkerg:
just wanted to correct an error, that's all.

vortexing:
And we appreciate it. ;) Just made me chuckle.

PR #646, opened 2020-11-11
Windows in Computing Resource Access

dirkpetersen:
Improved the Windows client docs in Computing Resource Access Methods, e.g. by adding WSL2 and a  VcXsrv X11 server

### DIFF
--- a/_scicomputing/access_methods.md
+++ b/_scicomputing/access_methods.md
@@ -32,9 +32,9 @@ This is a list of clients and servers by OS:
 
 ### Windows
 
-While puTTy is the most widely used secure shell software for Windows and you can easily install it through "Software Center" it is now a little dated. The most flexible way to get terminal access in Windows is using WSL2. WSL2 allows you to run Linux inside Windows and it is very fast (unlike classic virtual. machines). 
+While puTTy is the most widely used secure shell software for Windows and you can easily install it through your Fred Hutch device's "Software Center".  The most flexible way to get terminal access in Windows is using WSL2. WSL2 allows you to run Linux inside Windows and it is very fast (unlike classic virtual machines). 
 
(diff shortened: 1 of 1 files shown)


dirkpetersen:
I made a request to helpdesk for an installation of WSL2 and a VcXsrv but I did not add to the docs yet as it may take a while

bmcgough:
Does this really work? The steps I've seen are quite involved. I think the user who is running WSL(2) can probably figure this out on their own. And if opening access controls and firewall rules are a part of the config (which is what I have seen) do we want users who may not know what they are doing altering these rules?

dirkpetersen:
I installed this setup at home with these exact instructions . Took only a few min. WSL2 is not quite mainstream yet but I heard that folks in Center IT were using it.
For affiliates who work with multiple different sites+cloud this setup is better than nomachine

vortexing:
Soo.... do we have a security concern here?

dirkpetersen:
The last agreement was that Mike Nescot would be included in pull requests to review security. I think Kate has already a proposal.  Ultimately the center needs to decide if it wants to support WSL2. As to opening ports on local WS: This would be something you desktop support person would do if you do not have admin access and you do yourself if you do have admin access (few will have this in the future)

vortexing:
Does Mike have a github username to tag?

dtenenba:
@mnescot

dirkpetersen:
I have now found the perfect setup for myself VcXsrv with Winodows Terminal on WSL2 , works nicely with tmux and screen.  If there was a VcXsrv deployment through software center they could apply more custom / intelligent local windows firewall settings

vortexing:
@bmcgough What do you think we should do about this situation?

PR #651, opened 2020-11-19
some corrections and

dirkpetersen:
changes to the archiving docs


vortexing (commented):
Let's show people how to get a list of the subcommands and remove the man-page text from this page.  Just explain any particularly important subcommands and that's all.  

Then the rest is fine and we'll deal with removing it when it's no longer needed b/c YES, it already IS in a compdemo!!! DUH!!!!

PR #652, opened 2020-12-07
Free resources

cloudytechi:
I am reaching out to you because I have a Data Science training product that I think your audience will appreciate. This product might be worth mentioning in your blogs.

Regards,
Cloudytechi.


vortexing:
Sorry, we're only documenting what is already available at the Hutch.

PR #659, opened 2021-02-12
Mrg/gizmo sched

atombaby:
Two pages in this PR.  The "compute_accounts" page contains details about what cluster accounts are and what comes with those.  The second, "compute_job_scheduling" is a bit of a deep dive into job scheduling.

I don't think this needs to be in the sidebar, but it does need a link into one of the other pages.  Likely the access page and maybe one of the other gizmo pages.  Need some ideas here.

### DIFF
--- a/_scicomputing/compute_accounts.md
+++ b/_scicomputing/compute_accounts.md
@@ -6,11 +6,11 @@ primary_reviewer: atombaby
 
 The compute resources provided by Scientific Computing are available to all Hutch researchers. As a shared resource in high demand, it's necessary to manage usage to ensure some measure of equitable access to all research.
 
-This is primarily seen in the Gizmo cluster, where we have both a) limits on how much can be used by individuals and groups and b) algorithms that prioritize pending work.  The limits and prioritization algorithms serve to schedule the waiting jobs for available resources.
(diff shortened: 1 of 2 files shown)


bmcgough (approved):
Yes, this is great! In the future we should perhaps take an approach like..."Our cluster is co-funded by Fred Hutch and the NIH for all researchers on Fred Hutch. To this end we balance usage... ." This would be transparent, and remind everyone how/why this is a shared resource.

Also would like to include a description of over-subscription. This is an important concept to understand when looking at our scheduler and algorithms, and I think it is not widely understood.

lgreenen:
I agree with Ben's note the recommended statement and oversubscription.  

Should there be some statement about submitting a SciComp ticket if there is a need to have some other allocation for the researcher's work submission?  The reserving a number of cores for ad hoc short term use.

lgreenen (approved):

atombaby:
@lgreenen You're really whoopin' my ass on those restrictive clauses.. 8-)

> Should there be some statement about submitting a SciComp ticket if there is a need to have some other allocation for the researcher's work submission? 

Perhaps.  I can add that to the accounts discussion.

> The reserving a number of cores for ad hoc short term use.

We're not doing that currently.

lgreenen:
I have been called out in my emails by various researchers about my clarity – mostly from non-native English speaker.  They trained me.

I think that would be a statement to add but I am worried that every time someone runs a job, they will request all the ad hoc resources…

I still think we should add it.

Loren
Sr. Program Manager, SciComp

Issue #661, opened 2021-02-16
Genomics and bioinformatics analysis types


k8hertweck:
### URL of existing content (a rendered, sciwiki.fredhutch.org link)

Add resources explaining [types of services available from genomics/bioinformatics](https://www.fredhutch.org/en/research/shared-resources/core-facilities/genomics-bioinformatics.html). 
In particular, the [wiki genomics page](https://sciwiki.fredhutch.org/generation/datagen_platformsData/) includes information on [genomics core services](https://www.fredhutch.org/en/research/shared-resources/core-facilities/genomics-bioinformatics/genomics-services.html), like genomic platforms and some data types,
but we should include the rest present on the SR page.
Additionally, a parallel page for [bioinformatics standard services](https://www.fredhutch.org/en/research/shared-resources/core-facilities/genomics-bioinformatics/bioinformatics-services.html) would be awesome.

### Specific information requests or concerns, or specific region on page of content to refresh.

Each type of platform, data type, and analysis type should have links to additional resources. The wiki page already includes a "resources" section at the bottom, but it would be great to organize this information a bit better (e.g., what does each link provide?). Additionally, let's compile resources for each data platform/type and analysis, such as:
- video (YouTube, etc) explaining/showing technology
- primary research articles explaining technology/analysis
- blog posts, company links, etc 

### @ GitHub username of the listed primary reviewer of the page if known
@k8hertweck

vortexing:
This is now beyond anything we can realistically make happen.

PR #662, opened 2021-02-16
Genomics + bioinformatics resource dump

k8hertweck:
Addresses #661 

Not sure where this information will eventually end up, but for now, we want to be sure we have some high-quality reference materials for all of the standard services offered by Shared Resources.


k8hertweck:
Closing this since I've merged this branch into the `HDC-training` branch so it can be included in the addition of the HDC subdomain.

PR #671, opened 2021-04-22
Add instructions for use of restart QOS

atombaby:
Strangely lacking in our current documentation.  This is a change from how many are using it today but does reduce opportunities for errors. Currently, since the QOS plugin isn't compatible with Slurm 20+, it's necessary to use

```
sbatch -p restart-new --qos=restart-new ...
```

To use that partition.  This method documented in this PI eliminates the partition request and only sets the QOS (which is what makes the jobs restart-able).  The partition was necessary before to allow restart jobs to use private nodes, but currently there is just the single partition.


vortexing:
Nothing looks like it needs formatting adjustments.  @atombaby if you are ready to merge, change this from a draft to an official PR and we'll merge it in!!

PR #673, opened 2021-06-02
Adding Cellprofiler to the HDC workflow catalog

DevNambi:
Adding the CellProfiler workflow to the HDC workflow catalog


vortexing:
@DevNambi when you make new contributions just make new branches for them. We do try to delete branches when they are merged just for sanity's sake when everyone is in there.

vortexing (commented):
Can you make all the header levels for all the workflows in the catalog the same level?  I think they all need to be H3, as H2 is the top level header on these pages (H1 is reserved in the Wiki).

DevNambi:
I'll make a new branch & submit a new PR with the updates

PR #682, opened 2021-08-13
Added content for hosted reference genomes and automated builds for docker images

sminot:
I hope it's ok that this PR contains two different tranches of content. I've built this locally and tested the links, and so I think it should be pretty good. Thanks for reviewing!


sminot:
@bmcgough did you have more questions about this content?

vortexing (approved):
Let's do it!!  FINALLY someone wrote out what is IN /shared/biodata!!!!! It's only been 10+ years since I dreamt of this.

PR #687, opened 2021-09-30
Update access_methods.md

chrisequalsdev:
Updated section on transferring a ssh key to a remote host
Updated section on storing the key passphrase in the Mac keychain

### DIFF
--- a/_scicomputing/access_methods.md
+++ b/_scicomputing/access_methods.md
@@ -127,7 +127,6 @@ No one wants to type a long passphrase every time they use SSH. Creating an SSH 
 ##### MacOS Keychain
 The following steps will get your MacOS device into a state where your SSH key passphrase is unlocked for you when you log into the Mac, and is automatically supplied to SSH. You will need to be logged in to your Mac, have a terminal open, and have an SSH key to start.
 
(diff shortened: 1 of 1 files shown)


_scicomputing/access_methods.md
```diff
@@ -116,24 +116,25 @@ You will be prompted to enter and then confirm a passphrase. This is the best pr
...
+```
+On Windows it is easiest to cut-and-paste the key into a remote session opened using a password.
 #### Passphrase management
 No one wants to type a long passphrase every time they use SSH. Creating an SSH key with no passphrase is unsafe. There must be a better way... .
 
 ##### MacOS Keychain
-
 The following steps will get your MacOS device into a state where your SSH key passphrase is unlocked for you when you log into the Mac, and is automatically supplied to SSH. You will need to be logged in to your Mac, have a terminal open, and have an SSH key to start.
 
+* Change the permissions of your public key `chmod 600 <path to keyfile>` the default is `~/.ssh/id_rsa.pub`. MacOS will not let you add your 'key' to the keychain without doing this first.
```
atombaby, line 130: Does OSX need the public key to be mode 600?  Typically this can be ugo+r...

atombaby (approved):

_scicomputing/access_methods.md
chrisequalsdev: For some reason that wasn't working for me yesterday but is today. I will take this step out and leave permissions as default / 644 for the public key.

PR #695, opened 2021-11-22
update rModules

fizwit:
update R modules, improve section of .libPaths and what could go wrong with installing R libraries on the Rhino.

### DIFF
--- a/rModules.md
+++ b/rModules.md
@@ -14,11 +14,11 @@ toc: true
 toc_label: "On This Page"
 ---
 
-SciComp creates R modules for every version of R. There are two module for each version
-of R released. A base module **R/4.1.1** and a local Fred Hutch module **fhR/4.1.1**.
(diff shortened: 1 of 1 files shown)


dtenenba:
I made a few small changes (formatting mostly).

Do we want to put in a link to https://github.com/FredHutch/easybuild-life-sciences/issues/new to indicate where requests for new libraries should go? My only hesitation is that both the wiki and easybuild-life-sciences are public repos so this may result in noise or requests from non hutch people who don't quite get it....

fizwit:
if we had other documentation about how to open tickets, open issues and visit office hours. The we could just make a reference to center net or some other private source for how to open issues.

dtenenba:
Yeah, that would be good to have, but is beyond the scope of this PR right now, so I will approve it.

dtenenba (approved):
LGTM

PR #713, opened 2022-05-12
Update singularity instructions with notes on cache and tmp

atombaby:
Ran into a corner case with TMPDIR and image builds.  Used this as an impetus to document both of those environment variables.

Not terribly happy with the discussion of tmpdir, but might suffice for now?


vortexing (approved):
I'm not sure I understand the implication of your last statement that was IMPORTANT!  ???

vortexing (approved):
Let's do it!

PR #714, opened 2022-05-12
Update clsp_specimenBanking.md

sunilayala:
Please change the existing page to the content I've proposed. Lab Matrix is the old tool which has been discarded and we are going ahead with the Open Specimen implementation. 

(1) Please help to attach a PDF of the key features of Open Specimen where it says "Refer to the Key Features of the application here"

(2) Please help to format the Key Features section so that the features are shown in a grid (like how it is in the PDF) instead of one after the other


sunilayala:
@vortexing Please see if I've done it right this time - Thanks!

vortexing:
OK!  I added back in the information about specimen processing resource via Shared Resources b/c that's important for folks still.  I rendered the site locally and checked the table and fixed the youtube embedding and linked the pdf.  I think it's ready to merge!!!

vortexing (approved):
Let's get this in and then folks can triple check how it looks afterwards.

sunilayala:
@vortexing How do I delete the branch? When does the actual page on SciWiki get updated?

PR #715, opened 2022-05-16
Update guidance on configuring Nextflow for SLURM

sminot:
This reflects some guidance that has been helpful for Nextflow users on gizmo

### DIFF
--- a/_hdc/workflows/running/on_gizmo.md
+++ b/_hdc/workflows/running/on_gizmo.md
@@ -76,6 +76,8 @@ singularity {
 process {
     executor = 'slurm'
     queue = 'campus-new'
+    errorStrategy = 'retry'
+    maxRetries = 3
 }
 
 tower {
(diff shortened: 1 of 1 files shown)


_hdc/workflows/running/on_gizmo.md
```diff
@@ -91,3 +93,29 @@ If you have any problems using this configuration, please don't hesitate to
 
 To run a workflow with this configuration, follow the guidance for formatting
 the appropriate [run script](/hdc/workflows/running/run_script).
+
+## Note: ERROR 151
```
atombaby, line 97: I am always wary of asking people to ignore errors.  While I think retry is a good option in any case, the 151 error (really exit code 23) can also be indicative of problems that need to be addressed (e.g. I've seen this linked with memory issues on the node, where the job can be one of the contributors).

I do think this worthy of a note, but I'm not real enthused about including this in core docs...

sminot: The exit code in these cases are actually 0! Which is very perplexing, but not something that I've been able to figure out.

I share your reticence, which is why I didn't add it to the docs until now. But my experience has been that almost every person I know who is running Nextflow on gizmo has to end up adding this to their config at some point.

If it's any reassurance, any error which comes up consistently will get caught and reported to the user. It's not like a workflow is just going to ignore an error and keep on rolling through, which would be a very bad practice indeed.

PR #718, opened 2022-05-18
Overhaul shared resources

vortexing:
I reorganized the data gen sections and removed pages that had very little content in them but retained that by moving those bits into a Shared Resources page.  That allowed me to move the Specimen processing information out of the Specimen data management page but still keep it.


vortexing:
Gonna make an executive decision.

PR #719, opened 2022-05-18
Short partition and related content

atombaby:
Let's use this as a baseline for designing this partition.  I've got the name "short" still in place for the moment, and the limits & numbers are a bit of a wag.


vortexing:
Things!! 

- We should consider adding a table about partitions available and the sentiments behind them on this page:
https://github.com/FredHutch/wiki/blob/main/_scicomputing/compute_platforms.md

- Do we have a `restart` partition Resource library entry???  

- This page (https://github.com/FredHutch/wiki/blob/main/_scicomputing/compute_jobs.md) has too many headers so the TOC actually gets too long.  Consider simplifying it as part of this PR?  

- We should reorganize the Parallel computing page content in addition to this PR so that we streamline these docs a bit.  I feel like reading the parallel computing page should happen first prior to deciding on a partition in your job.  https://github.com/FredHutch/wiki/blob/main/_scicomputing/compute_parallel.md 

I don't mean to make your focused PR into a larger monstrosity, so feel free to ignore these, but I wonder if clarifying how to best leverage the cluster/partitions/slurm/workflow managers might be done all together to reduce content overall but also ensure that it's consistent across pages and in a logical order.  Maybe some portion of these thoughts could get integrated into this PR?

atombaby:
> adding a table about partitions available and the sentiments behind them on this page

I wasn't sure how to fold partition information into the platforms page- it seems a bit esoteric for an overview page such as this one.  The compute jobs page seems more logical, but as you've indicated, its overlong.  For the moment I've recorded that information here.

https://github.com/FredHutch/wiki/blob/mrg/short/_scicomputing/compute_partitions.md

atombaby:
Punchlist:

 - move into compdemos to link into resources page
 - update links in compute_jobs

vortexing (commented):
The `compute_partitions.md` file will get rendered but not linked to anywhere.  Should it also be moved to _compdemos and a link placed in the `compute_jobs.md`?  Or do we want to put `compute_partitions.md` in the sidebar on the SciComp domain?

atombaby:
I've updated and rearranged things so that the compdemos is the jumping off point.  We'll move this all into the scicomputing section when we redo `compute_jobs.md`

Ok to merge now?

vortexing:
compute_restart.md needs to go into compdemos and be linked to there.  

(I'm now also seeing more files in the _scicomputing directory that likely aren't being linked to by the sidebar that maybe should be in compdemos too... eeek! another PR)

Otherwise, it sounds like a great plan!!!

dtenenba (approved):
LGTM

PR #725, opened 2022-05-31
committing changes to aws-on-wsl branch

staciefase:
adding documentation on how to install and use AWS CLI on Windows Subsystem for Linux


atombaby:
I'd suggest we move this into the compdemos directory.  gendemos is a little more focused on the actual creation of the data itself.

staciefase:
> I'd suggest we move this into the compdemos directory. gendemos is a little more focused on the actual creation of the data itself.

Got it. I'm not familiar with the process here....is that something you do or something you'd like me to do and then submit another request? Are there other pieces of feedback I should fold in and just make all changes at once? Thanks!

atombaby (approved):

PR #728, opened 2022-06-08
Add instructions and infrastructure for building on rhino

atombaby:
So my workflow is easiest when I build this on a rhino/gizmo type system for testing.  Works well with the Ruby installed in app but a couple changes make this much easier:

 - adding `.bundle/config` to install gems in _vendor
 - adding the _vendor directory to gitignore
 - adding some instructions

I'm only concerned that the bundle config will break the build- does anyone have any insights on that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ To contribute to the Wiki you only need to have your GitHub username added to th
 
 [Building a copy of this Wiki locally](#building-the-site-locally)
 
+[Building a copy of this Wiki on a rhino](#building-the-site-on-rhino)
+
 [For Admins](#for-admins)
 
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -224,6 +223,40 @@ If you are inside the Fred Hutch network, you can type
...
+### Build
+
+```
+bundle exec jekyll build
+```
+
+### Serve
+
+```
+bundle exec jekyll serve -H <node IP address> -P <port>
```
dtenenba, line 251: You can use the node hostname (not just IP) or `0.0.0.0`.
Also note that this command will do a build as well, so you can skip the previous (`bundle exec jekyll build`) command if you want.

dtenenba (approved):
LGTM

PR #729, opened 2022-06-08
Positioning of desktop computing

vortexing:
Added a link to the other document that helps people think about where they need to run their code.


atombaby (approved):
My only concern is that Anaconda isn't a real great experience on rhino/gizmo so I hesitate to recommend it.  But it's still the gold standard for desktop installs (just so buttery-easy to install and manage)

vortexing:
I don't python, so please change it if you notice things!!!

PR #731, opened 2022-06-13
Deploy all branches

dtenenba:
As [documented](https://github.com/FredHutch/wiki/blob/deploy-all-branches/README.md#automated-deployment) in the README in this PR,

Everything merged into the `main` branch
will be automatically deployed to 
[https://sciwiki.fredhutch.org](https://sciwiki.fredhutch.org).

Everything in any *other* branch pushed to GitHub will be deployed to 
[https://sciwiki-preview.fredhutch.org](https://sciwiki-preview.fredhutch.org),
which is only accessible inside the Fred Hutch network.
You can check what branch and what commit is reflected by going to 
[https://sciwiki-preview.fredhutch.org/info.txt](https://sciwiki-preview.fredhutch.org/info.txt).

This is actually already live (so check it out), because it does not need to be merged into main in order to take effect. When I pushed these commits it triggered the functionality. When it is merged into main, nothing new will happen because this functionality only happens with branches that are not main.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -29,7 +29,7 @@ deploy_preview:
     - docker tag dockerimages.fhcrc.org/wiki:latest dockerimages.fhcrc.org/wiki-preview:latest 
     - docker push dockerimages.fhcrc.org/wiki-preview:latest
     - sleep 15
-    - rancher-v0.6.2/rancher --url https://ponderosa.fhcrc.org --access-key $PREVIEW_RANCHERAPI_KEY --secret-key $PREVIEW_RANCHERAPI_SECRET up -d --pull --force-upgrade --confirm-upgrade --stack wiki --file docker-compose-preview.yml --rancher-file rancher-compose.yml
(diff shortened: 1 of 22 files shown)


dtenenba:
When this is merged, please choose Squash and Merge instead of "Merge Pull Request" so that all my sloppy commits are squashed into one.

dtenenba:
@atombaby yes, I thought about ways to have multiple builds for multiple branches but it became very complex very quickly. This is why I added the `info.txt` which tracks which branch/commit you are seeing. If you have suggestions for how to implement multiple previews for different branches I would welcome them.

atombaby:
Ok- cool.  Just making sure.

Yes- building each branch could get hairy very quickly- I was mostly interested if this process would subvert the need for the "build on rhino" stuff I was working on.

Would it be easy enough to change the build process to build a preview branch?  Maybe get all of us out of the habit of merging/committing on main?

dtenenba:
That is possible, but I'm not 100% sure that it's a better solution. it's just another thing to remember to do.  And then what if two people push to the preview branch within minutes of each other? Then you're back to square one. I'd hope this would still be useful as is, and would obviate the need for building on rhino. Typically you would look at the preview soon after your push and there isn't usually enough pushing to overwrite the preview before you've had a chance to look at it.

Another thing we could consider is creating build artifacts - zips of the built site that you could easily view with `python3 -m http.server`. They could have names that incorporate the commit sha and the branch. This would be in addition to the preview site. What do you think?

atombaby:
Good point.  Nah, let's not complicate things further... I think most who want a "private preview" for their own work are using docker, and using `bundle exec jekyll serve` also works well for deploying a private preview of the site.

atombaby (approved):

PR #732, opened 2022-06-23
Fix broken link

sminot:
Found a link that was broken


dtenenba (approved):
Good spotting. 
Reminds me that there was some code that was supposed to check for broken links in the CI/CD process, I'll need to check on that and see if it can be resurrected.

PR #735, opened 2022-07-07
Clarify that bucket name field should be left blank in swift cloud co…

dtenenba:
…nnection

...believe it or not, someone put **Unnecessary** into this field.


vortexing (approved):
no comment.

PR #739, opened 2022-07-08
Check for orphan assets

### LINKED ISSUE #125: Check for orphaned images in assets folders
At some point we will benefit from cleaning a bit by identifying images in asset folders that are not specifically included in any markdown.

An added bonus would be to identify images in asset folders which are NOT inside the individual asset folder where the markdown that uses that image is so that we could uncross the streams!

dtenenba:
Closes #125


vortexing:
Is this just removing orphans or does it set up some sort of regular check?  Can I merge it without breaking something?

dtenenba:
it does not remove orphans, just fails the build if there are any orphans and tells you (if you look at the build report on gitlab) what the orphans are. I have removed the current orphans so if you merge it it will not break anything (not now anyway). 
It does not file an issue (yet). That should not be too hard to add.

vortexing (approved):
Cool.  I like it.  It should be a good check to add even w/o the issues for now.

dtenenba:
Thanks. I'll do another PR when I have the issue thing done, should not take long.

PR #745, opened 2022-07-11
File issue re orphans

dtenenba:
Files an issue if there are orphan assets.


vortexing (approved):
yessss.   let's do it!

PR #750, opened 2022-07-14
edits to 'build site on rhino' process

dtenenba:
External IP of node not needed - use 0.0.0.0 instead. 
Other minor tweaks.


vortexing (approved):
sounds good.

PR #756, opened 2022-07-21
DON'T MERGE THIS YET - Update to ruby 3.0.1, eliminates warnings

dtenenba:
DON'T WANT TO MERGE straightaway, let's discuss first.

The goal of this PR is to eliminate the copious warnings that show up when building the site:

```
/home/dtenenba/dev/wiki/_vendor/bundle/ruby/2.7.0/gems/jekyll-3.8.5/lib/jekyll/document.rb:449: warning: Using the last argument as keyword parameters is deprecated
...repeats ad nauseam...
```

Deprecation warnings often turn into errors in later versions of software, so it's best not to ignore them.
To that end, this PR updates Ruby to 3.0.1 (not the latest version available, but the latest for which there is an Easybuild recipe ready to go, so the latest available on the rhinos/gizmos). 

Also, version pins are removed from Gemfile. I am not sure about the wisdom of leaving it like this. I removed them so that we would get the latest versions of all packages. But now that I've done that, perhaps we should re-pin each package to its current version so no breaking changes or regressions can bite us. I may add a commit that re-pins the package versions.

I have tested this (and it works) with:

* GitLab CI build
* Docker build instructions (in README)
* Rhino build instructions (in README)

I have *not* tested it with the "local" build instructions in README because I've had trouble installing Ruby-3.0.1 on my local machines (for different reasons). 
So one more reason we may want to hesitate on this.

At any rate, if those warnings ever do turn into errors, merging this ought to fix the problem.

Reasons we might want to have a discussion before merging this are:

* People may want to make sure they can set up a local Ruby 3.0.1-based build setup before we make it the default.
* This is taking us further down the road away from the way GitHub Pages does things. If we ever want to go back to GitHub pages we'll need to change more things back. I don't think we plan to do that, but it's worth mentioning. The original motivation to move away from Pages was to be able to use plugins (like `jekyll-glossary_tooltip`) that are not supported there, and I don't think we want to give that up.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # build me as fredhutch/wiki
 
-FROM ruby:2.7.1
+FROM ruby:3.0.1
 
 # FROM ubuntu:18.04
 
--- a/Gemfile
+++ b/Gemfile
@@ -5,15 +5,18 @@ group :jekyll_plugins do
 end
 group :development, :test do
   gem "github-pages", group: :jekyll_plugins
-  gem 'jekyll-redirect-from', '~> 0.14.0'
-  gem 'jekyll-sitemap', '~> 1.2.0'
-  gem "nokogiri", ">= 1.10.4"
-  gem 'jekyll-toc', '~> 0.6.0'
+  gem 'jekyll'
+  gem 'jekyll-redirect-from'
+  gem 'jekyll-sitemap'
+  gem "nokogiri"
+  gem 'jekyll-toc'
   gem 'jekyll-include-cache'
-  gem 'jekyll-gist', '~> 1.5.0'
-  gem 'jekyll-remote-theme', '~> 0.3.0'
-  gem 'jemoji', '~> 0.10.0'
-  gem 'html-proofer', '~> 3.14.0'
-  gem 'rake', '~> 12.3.0'
-  gem 'faraday', '~> 0.17'
+  gem 'jekyll-gist'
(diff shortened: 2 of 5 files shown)


vortexing:
I think we're pretty far from vanilla GitHub pages, so not to worry too much about that.  But I will try to build this locally and see if I can do enough testing to provide a check that "it works".

vortexing:
@dtenenba I have tested this locally and it works as written.  I have also filed an issue re: gem version pinning.  This could be merged.

vortexing (approved):
Tested and it works.

dtenenba:
Added a commit that pins the version number of the gems. So that should futureproof things until the next time we need/want to upgrade something. 

So now the only hesitation is that not everyone might have Ruby 3. I had trouble installing it on a couple machines. But if I spent more time on it I think I could get it to work. And there are always the docker and rhino build methods.

vortexing:
Well I think the local build process now HAS to be the docker build option though.  I guess I misunderstood.  Didn't we break the actual jekyll build process locally long ago?

vortexing:
I'm testing this for real now.  Sorry about my brain fart. ;)

dtenenba:
No, we did not break the local build process before.

vortexing:
Very sorry.  The local build process WITHOUT docker also works on my computer at least.  Sorry for that hiccup!!

vortexing:
closes issue #761

PR #778, opened 2022-08-16
Announcement for Maintenance Work

atombaby:
Discuss- anything more we need to do?  Is this too much?

### DIFF
--- a/_scicompannounce/2022-09-10-rhino-gizmo-updates.md
+++ b/_scicompannounce/2022-09-10-rhino-gizmo-updates.md
@@ -61,26 +61,3 @@ This is an entirely back-end change that won't impact or change use of Slurm ser
 
 Lastly, everything will get a reboot.  This is required to update the kernel, but has the added benefit of "clearing out the cobwebs" that can accumulate on systems that are running for long periods of time at the high loads.
 
-
-
-
-
-
-
- -----
-The gizmo cluster will be unavailable during the Scratch file system rebuild.   To prevent jobs from running during this time we have placed a maintenance reservation on gizmo.
-
-
-## Updates to the Gizmo Environment
-
-We are making a few changes to the gizmo/rhino computing environment.  These are described below.  Contact Scientific Computing by filing a helpdesk ticket (`scicomp@`) if you need assistance or have further questions.
-
-### Slurm Upgrade
-
-We'll be updating Slurm to version 21.08.  There are a number of changes which we don't expect will impact our current use of Slurm but upgrading will keep us well positioned for support.
-
-Detailed changes are available in the [release notes](https://github.com/SchedMD/slurm/blob/slurm-21-08-0-1/RELEASE_NOTES).
-
-### Move app file system
-
-We're moving the `/app` file system (hosting the software modules) to a different file server in anticipation of the retirement of the server it's currently running on.  This should not change how you use modules.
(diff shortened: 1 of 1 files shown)


_scicompannounce/2022-09-10-rhino-gizmo-updates.md
```diff
@@ -0,0 +1,63 @@
...
+```
+or
+```
+sbatch --time=12:00:00 # Requests 12 hours
+```
+
+if the job has already been submitted, update using `scontrol`:
+
+```
+scontrol update jobid=<jobid> timelimit=1-0
```
dtenenba, line 29: Maybe explain what this time limit is?

_scicompannounce/2022-09-10-rhino-gizmo-updates.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+## The Work
+
+### Kernel Upgrade
+
+We'll be updating the kernel (and associated libraries) on rhino/gizmo nodes to the latese maintenance release from Canonical.  These are primarily bug and security fixes- we don't expect any changes to arise from this.
```
dtenenba, line 36: typo: latese

_scicompannounce/2022-09-10-rhino-gizmo-updates.md
```diff
@@ -0,0 +1,63 @@
...
+
+While this is a rolling upgrade which doesn't affect many services, the nature of our HPC work has in past resulted in problems as individual nodes in the _silver_ cluster are rebooted.  Thus, we've decided to take a full outage and to use this as an opportunity to perform some much-needed maintenance on those systems.
+
+This document describes the scope of the outage, the updates we're making, and what you can expect in the lead-up to the outage.
+
+## What to Expect
+
+During this maintenance the rhino login nodes and gizmo compute nodes will be unavailable for any work.  Attempts to connect via ssh will fail.  Any processes running on these nodes will be terminated.
+
+As we have done in the past, we have put in a reservation against all cluster nodes starting at midnight of the 10th. Jobs that would not complete before the maintenance will start will be held until after the work is complete.  `squeue` will show the job as pending with reason _ReqNodeNotAvail, Reserved for maintenance_.
```
dtenenba, line 15: This might require some clarification. If a job is running before the maintenance, it will not continue running afterwards, will it? If so I would expect that it would not produce the expected results. 
I think what you are saying only applies to jobs that have not started yet, right?

dtenenba (changes requested):
See inline comments.

_scicompannounce/2022-09-10-rhino-gizmo-updates.md
atombaby: If a job is running before the maintenance, then the time requested would indicate that it would finish before the maintenance.  If it goes over time, Slurm will kill it (after 72 hours anyway).

I've _never_ been able to do a real good job explaining this.

dtenenba: OK, I think I get it. Others probably won't be quite as dim as me....

atombaby:
Ok, updated and added an example describing time limits and how they're impacted by the reservation.

dtenenba:
That is great. Very clear, thanks.

dtenenba (approved):
LGTM

PR #781, opened 2022-08-19
A few more details about setting up globus connect

atombaby:
I've been debugging a problem with globus and have picked up a few tips what might be useful...


vortexing (approved):
Let's do it!

PR #784, opened 2022-08-29
Use local elasticsearch

dtenenba:
This PR changes the way the site search works. Currently search queries go to an AWS Elasticsearch cluster in the old SciComp AWS account.

That account is going to be decommissioned soon. 

I looked at migrating this Elasticsearch cluster to the new SciComp AWS account and I ran into some problems. I may continue to investigate those problems, but along the way I had the idea that, now that we are hosting the wiki in our CI/CD platform (not GitHub pages), it would be possible to host our own single-node Elasticsearch cluster in a container that runs along with the wiki container. So that's what this is.

So in this PR, the indexing of the site is part of the site building process. It's not done anymore by an AWS Lambda function (that lambda function will need to be decommissioned, but the AWS account is going away anyway). 

The nginx configuration of the wiki container was modified slightly, to proxy requests to the Elasticsearch container. Only requests that start with `/sciwiki0/_search` are proxied (`sciwiki0` is the name of the ES index and `_search` is the endpoint for searching). This means no other endpoints can be accessed from the outside. This prevents users from doing anything with our ES cluster except searching it (for additional security, only GET and POST requests are proxied to ES). It also means we don't have to worry about CORS at all because all requests are going to the same host/port. And we don't have to pay for the ES cluster in AWS anymore.

Note that when you push to a branch other than `main`, the search index is built and available at 
https://sciwiki-preview.fredhutch.org (inside the FH network). This is separate from the index on the production site. 
So if you make a branch with a page that has to do with peaches, you can search for peaches in https://sciwiki-preview.fredhutch.org and the appropriate results will show up. When you merge your branch into `main`, then the same search will work on production.

One slight change is that I've changed the way pages are indexed. The current way is that a crawler (running inside a Lambda function) actually makes http requests, starting with the index page, and follows links until it has hit every page. 
Since the new method runs in our CI/CD pipeline, we actually have access to the freshly created pages, so we just open them as files on the filesystem. This is faster and avoids all the web traffic associated with crawling the site.
One slight downside is that we may end up indexing orphan pages (pages which don't have any links to them), since we are just indexing all `index.html` files, and not following links. 

I don't think we have a lot of orphan pages but I'll open an issue to track this and come up with a solution. I think there are some tools in the Jekyll ecosystem to handle this. A quick search shows that there is a [GitHub action](https://github.com/MOLOCH-dev/jekyll-orphan-pages-action/blob/main/action.yml) to do this that could easily be ported to run in GitLab or just as a cron job.

If you merge this PR, please choose the option to squash commits if it is possible (it may not be, because somehow a merge happened at some point). This ended up being very fiddly and there are tons of silly commits.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.gitignore
+tmp/
+*.gz
+rancher-v0.6.2
+elasticsearch/
+.gitlab-ci.yml
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@
 .vscode/
 _site/
 .idea/*
-
 Gemfile.lock
 *.code-workspace
 _vendor/*
+info.txt
+tmp/
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -45,7 +45,7 @@ build_test:
     docker network create wiki-indexer || true
     # run an instance of elasticsearch
     docker rm -f elasticsearch || true
-    docker run --network wiki-indexer -d --name elasticsearch -p 9200:9200 -e "discovery.type=single-node" -e "xpack.security.enabled=false" elasticsearch:8.4.0
+    docker run --network wiki-indexer -d --rm --name elasticsearch -p 9200:9200 -e "discovery.type=single-node" -e "xpack.security.enabled=false" elasticsearch:8.4.0
     sleep 20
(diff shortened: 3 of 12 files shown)


vortexing (approved):
let's do it!!

PR #800, opened 2022-10-03
plumbing changes

dtenenba:
A bunch of infrastructural changes - no content changes. 

For details about this PR, please see this Slack post:

https://fhdata.slack.com/archives/CA2KRB666/p1664829419189939

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ Gemfile.lock
 _vendor/*
 info.txt
 tmp/
+_pythonModules/*
+_rModules/*
+easybuild/*
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,6 +74,7 @@ create_search_index_0:
     set -x
     # build search index
     # run image and copy HTML files out of it
+    docker rm -f wiki || true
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -7,4 +7,7 @@ Gemfile.lock
 *.code-workspace
 _vendor/*
 info.txt
-tmp/
\ No newline at end of file
+tmp/
+_pythonModules/*
```
vortexing, line 11: Do we need to include _pythonModules and _RModules?  or not b/c they are old and thus re-pulled during every PR from the easybuild repo???

vortexing (approved):
Assuming the _Modules folders should be in the gitignore, then LGTM!

.gitignore
dtenenba: They are no longer part of this repo - they get pulled from the easybuild repo during the build process. 
It's not so relevant anymore but when I would troubleshoot builds on the shell runner I would do 'git status' and it would show me hundreds of files from these two directories because they were not in .gitignore and not in git either. 
That's why I added them to .gitignore.

Issue #804, opened 2022-10-12
The new style changes make inline code blocks hard to see


chrisequalsdev:
**URL of page(s) of concern**
Specify the URLs of the page(s) where there are technical problems.
https://sciwiki.fredhutch.org/scicomputing/store_posix/#how-to-access-fred-hutch-storage

**Describe the problem**
Inline code blocks should be shaded darker as they are difficult to discern from regular text

chrisequalsdev:
**Desired fix:**
#fafafa -> #afb8c133 (this is what github uses for inline code blocks)

PR #809, opened 2022-10-19
Update main.scss

chrisequalsdev:
CSS change to format inline code blocks with a darker background (matches what github uses) to make them easier to see. I tested this locally.

Resolves: #804


dtenenba (approved):
👍

Issue #808, opened 2022-10-19
Edit this page via github links to master branch instead of main


chrisequalsdev:
**URL of page(s) of concern**
https://sciwiki.fredhutch.org/scicomputing/comp_index/

**Describe the problem**
Link is to master branch, should be main. It redirects but link should be fixed

PR #810, opened 2022-10-20
Update social-share.html

chrisequalsdev:
Change repository url to main in template. With this change, all pages now link to main instead of master. I tested this change locally.

Resolves: #808


dtenenba (approved):
Good catch.

PR #815, opened 2022-10-26
Suggestions on reformatting pathway for economy local

atombaby:


vortexing (approved):
I think this is more of the format of things that we'd all intended with the Pathways.  There are some helpful tips in @jnyhuis 's content that coudl be put directly into the Object Store or Motuz pages for folks, but this pathway is more for introduction/context of those pages which is great.

PR #820, opened 2022-11-11
moving/relinking privacy/security/legal

vortexing:
This content is the first set of info that given our new HIPAA hybrid situation will be important to update and refine.  Also it relates to data sharing issues as well.  

DaSL, do you have any edits and wisdom you can provide or content you can borrow (so its full text searchable IN the wiki), and/or links you can link out to?  

My aim is to get some ISO and BD and IRB input on these but that shouldn't hold up the PR.  This content is ALREADY in the wiki but in a place we are deprecating.  So this PR aims to bring it into more focus in the Data Science section so we can better request input on it from the community.


carriewright11:
This looks pretty good to me! I am working on a course on these topics but it isn't quite done yet - it needs some reorganizing mostly and some spots filled out. Feel free to take a look if there is anything you are interested in using or thinking about adding. When the course is done you are welcome to link out to it- I shifted gears to something else and will return to developing this course later: https://jhudatascience.org/Ethical_Data_Handling_for_Cancer_Research/

PR #824, opened 2022-11-20
add two more courses

vortexing:
Thought it'd be good to mention these two other courses/guides in this table too.


carriewright11 (approved):
LGTM!

Issue #828, opened 2022-12-02
`bundle install` fails with conflicts on Gem versions


atombaby:
**URL of page(s) of concern**
 
N/A

**Describe the problem**

The version pin for faraday (~> 2.3.0) seems to conflict with jekyll-gist:

```
rhino03[~/Work/wiki](mrg/nodetypes): bundle install
Fetching gem metadata from http://rubygems.org/.........
Resolving dependencies...
Bundler could not find compatible versions for gem "faraday":
  In Gemfile:
    faraday (~> 2.3.0)

    jekyll-gist (~> 1.5.0) was resolved to 1.5.0, which depends on
      octokit (~> 4.2) was resolved to 4.22.0, which depends on
        sawyer (~> 0.8.0, >= 0.5.3) was resolved to 0.8.2, which depends on
          faraday (> 0.8, < 2.0)
```

PR #829, opened 2022-12-03
Remove version pin for faraday

atombaby:
Resolves issue #828 .  I'm not sure if a version pin for this is required- dependencies seem to resolve OK and I can build and serve with bundle exec, so it seems to work OK


dtenenba:
See #830 for a different approach.

atombaby:
Abandoned in favor of #830

Issue #769, opened 2022-07-29
Consider moving some of the Bionic upgrade updates to announcements section instead to archive?


vortexing:
Or remove them if they are no longer relevant.   Or leave them if they are fine. 

### URL of existing content (a rendered, sciwiki.fredhutch.org link)
https://sciwiki.fredhutch.org/compdemos/gizmo-bionic-changelog/
https://sciwiki.fredhutch.org/compdemos/gizmo-bionic-known_issues/
https://sciwiki.fredhutch.org/compdemos/gizmo-bionic-upgrade/
https://sciwiki.fredhutch.org/compdemos/gizmo-nx-users/

### @ GitHub username of the listed primary reviewer of the page if known
SciComp crew

vortexing:
Discussion 1/13/23:
1.  Consider moving this section of text: https://sciwiki.fredhutch.org/compdemos/gizmo-bionic-upgrade/#software-and-environment-modules-lmod into https://sciwiki.fredhutch.org/scicomputing/compute_environments/#environment-modules
2. Then move the gizmo-bionic-upgrade page itself over to the scicompannouce directory.  
3.  Then check for broken links

vortexing:
Consider 
1.  moving https://sciwiki.fredhutch.org/compdemos/gizmo-bionic-changelog/ to scicompannounce as is, 
2. then check for broken links

vortexing:
Consider:
1.  moving https://sciwiki.fredhutch.org/compdemos/gizmo-bionic-known_issues/#binary-incompatibility-between-node-classes section over to https://sciwiki.fredhutch.org/scicomputing/compute_platforms/
2.  Move the rest of that page to scicompannounce
3.  check for broken links.

vortexing:
Consider:
1. combine content from https://sciwiki.fredhutch.org/compdemos/gizmo-nx-users/ into https://sciwiki.fredhutch.org/scicomputing/access_nx/

atombaby:
I've moved all of the indicated pages.  Is it essential to update links in there?  The information is nearly 2 years old now...

atombaby:
I went over the info in those pages- the only thing I located that really needed to be added to the main docset is some information about the alternate NX session methods (see [c2cd0e1](https://github.com/FredHutch/wiki/commit/c2cd0e187dae95f0136918dd7b89e3a01fe790ff))

I've created PR #840

atombaby:
Fixed in #840

PR #840, opened 2023-01-19
reorganize bionic announcements

atombaby:
move the announcements in compdemos relating to the bionic update in 2020 over into the announce section.  See issue #769


vortexing (approved):
closes #769  as long as @dtenenba also feels like what material that should be in main pages is there.  Were there any other gems that you'd thought to retain?  It's not like anything is going away of course.

dtenenba:
👍

PR #845, opened 2023-01-27
Update information about secure vs. fast file

atombaby:


vortexing (commented):
No one really knows how to contact ISO, so more specific instructions would be better.

atombaby:
ALSO needs information about thorium

atombaby:
So I actually decided to pull out that bit about contacting iso for fast file PHI approval.  It's not clear to me what we're having the reader _ask_ iso about phi and fast file.

I've also added more details on scratch and onedrive.

vortexing (approved):

PR #850, opened 2023-03-08
Add snippet for adjusting nice value for pending jobs

atombaby:
Instructions on setting individual user priority were missing instructions to adjust nice value on queued (pending) jobs.


dtenenba (approved):
LGTM

PR #852, opened 2023-03-10
Update FAQ question on additional S3 buckets

atombaby:
Add additional information about charges and process


dtenenba (approved):
Perfect.

PR #857, opened 2023-03-24
Clarify information about filesystem access permissions

atombaby:
There was an incomplete discussion of file system permissions- this has been expanded slightly with the scope reduced to just fast-file.  The S3 access permissions information was obsolete and removed.

This includes information on collaboration directories and navigating blind parent directories


vortexing (approved):
I think this will help!

PR #869, opened 2023-04-14
Singularity to apptainer

dtenenba:
Change references to Singularity to Apptainer

Note that this is not a wholesale search and replace.
In each case I made sure that the apptainer version
of commands, environment variables, etc., were valid. I put "(formerly Singularity)" at least once on each page that mentions Apptainer, so that the old name can still be found by searching.


_compdemos/Apptainer.md
```diff
@@ -0,0 +1,265 @@
+---
+title: Using Apptainer Containers
+last_modified_at: 2021-07-30
```
sminot, line 3: Probably should update this

atombaby (approved):

_compdemos/Apptainer.md
dtenenba: This gets automatically updated by a lambda function.

PR #873, opened 2023-05-03
Data Core: Add content on Cirro and office hours

sminot:


sminot:
I'm not totally sure that I added the image in the right way. My apologies, I wasn't able to get my local testing environment up and running before putting in this PR.

PR #875, opened 2023-05-24
Add docs on lmod defaults

atombaby:
So a ticket came in the other day- raises a bunch of questions about how we handle module defaults.  Led me down a rabbit hole where I found this information.

I'm being a little opinionated- I'd like to be a little _more_ opinionated, but I'm not sure how to work that in.  Anyway, LMK whatcha think


dtenenba (approved):
I like it.

PR #876, opened 2023-05-24
Short TOC for resources that new people will find useful

atombaby:
As per a conversation we had today- this doc will identify resources that a new research community member could use to get started with computation at the Hutch.

This identifies starting points for getting help and information.

This will be sent as an email to our scientific relationship managers and research admins  so that they can, in turn, give this to new hires in their groups.


vortexing (approved):
Did a wee tweak!  Let's do this.

PR #885, opened 2023-07-05
update docs for using putty on Win

chrisequalsdev:
Add some images and better detail on connecting to rhino using PuTTY


dtenenba (approved):
Looks good.

PR #892, opened 2023-07-18
Deidentify methods

monicagerber:
Tidyverse instructions for de-identifying data.

Eventually will have methods for unstructured text.

@vortexing could we start a resource library for this?


monicagerber:
@vortexing I think I fixed it by changing to GitHub-flavored-markdown style code chunks.

atombaby (approved):

PR #896, opened 2023-07-31
Use jupyter with r

dtenenba:
After getting two requests for this in two days, it seemed like we needed some documentation for it.

One thing I'm not sure about is whether to tell people to use rhino, or should we have them use a gizmo node?
And if so, how does that work with nomachine? 
If I run grabnode inside a NoMachine terminal, I end up without X11 so the notebook does not run.

Does NoMachine run on gizmo nodes, or just on rhinos?

### DIFF
--- a/_data/glossary.yml
+++ b/_data/glossary.yml
@@ -45,5 +45,5 @@
   definition: Singularity is the original name for the Apptainer project
 - term: Apptainer
   definition: Apptainer is a system for using containers in an HPC environment
-- term: major version of R
-  definition: All versions of R that share the same `y` of the `x.y.z` version number. For example, R-4.2.1 and R-4.2.2 share the same major version number. R-4.1.1 and R-4.2.2 do not.
+- term: minor version of R
+  definition: All versions of R that share the same `y` of the `x.y.z` version number. For example, R-4.2.1 and R-4.2.2 share the same minor version number. R-4.1.1 and R-4.2.2 do not.
--- a/_data/glossary.yml
+++ b/_data/glossary.yml
@@ -45,3 +45,5 @@
   definition: Singularity is the original name for the Apptainer project
 - term: Apptainer
   definition: Apptainer is a system for using containers in an HPC environment
+- term: major version of R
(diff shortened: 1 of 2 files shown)


atombaby:
> If I run grabnode inside a NoMachine terminal, I end up without X11 so the notebook does not run.

What's the error you're getting there?

> Does NoMachine run on gizmo nodes, or just on rhinos?

It only runs on rhino nodes.  We should be getting a display in Slurm allocated sessions- I am getting an "Unable to open display" error, which is odd because I thought we had that working...

dtenenba:
This is the error:

```
[I 13:43:10.670 NotebookApp] Kernel started: e16dcf6b-e24f-47c4-9d18-ca738a20070b, name: ir
Error in .External2(C_X11, paste0("png::", filename), g$width, g$height,  : 
  unable to start device PNG
Calls: <Anonymous> ... evaluate -> dev.new -> do.call -> <Anonymous> -> ok_device
In addition: Warning message:
In ok_device(filename, ...) : unable to open connection to X11 display ''
Execution halted
```

atombaby:
Try `grabnode --x11` and see if that changes anything...

dtenenba:
That works! 

OK, I guess that's what we should tell people to do - we don't want them cluttering up the rhinos with potentially heavy duty notebooks. I'll update the PR...

atombaby:
I'm thinking we should make that the default for grabnode.  I thought it was, not sure if that's a regression

dtenenba:
Yeah, that sounds good to me. Might want to add a --no-x11 flag just in case you don't want it, but I can't currently imagine a use case for that.

dtenenba:
OK, I have added the documentation for the grabnode part.

This feels a lot like the way we used to do rstudio server before there was a web app. Lots of steps and kind of clunky, making non-command-line folks do command-line stuff.

Should just be a one-clicker. Open OnDemand is certainly the solution here, but if that is going to take a long time, we could probably make a 'jupyter lab/notebook launcher' application like the rstudio launcher.

atombaby (approved):

PR #899, opened 2023-08-15
X11/NoMachine dependency is not needed for Jupyter+R

dtenenba:
It turns out that @fizwit was right and we just need to set the `bitmapType = 'cairo'` option to get around the need for X11/NoMachine when running Jupyter Lab + R. So this PR documents that (simpler) process.

Also changed some verbiage to discourage running Jupyter directly on rhinos and encourage doing so in a grabnode session. I'll make another PR to change the similar verbiage on the Python/Jupyter page.


chrisequalsdev:
Looks good.

chrisequalsdev (approved):
Looks good to merge.

Issue #886, opened 2023-07-05
faster-dump Issue Page


fizwit:
### fasterq-dump Issue
fasterq-dump breaks when used on scratch. The only fix is to reboot the host which is not good.  Possibl fix is to just remove it.  Update page to recommend using the module `parallel-fastq-dump/0.6.7-GCCcore-11.2.0`

### @atombaby reviewer

See issue: https://github.com/ncbi/sra-tools/issues/383

PR #907, opened 2023-09-01
Update fasterq dump workaround instructions

atombaby:
Include use of `parallel-fastq` as a workaround.  Addresses #886


fizwit (approved):
lgtm

PR #922, opened 2023-10-09
Update Cirro documentation

sminot:
Provides a description of how the Cirro CLI module can be used from the rhino/gizmo to access data hosted in Cirro, along with some pointers into other documentation which may be useful.


dtenenba (approved):
This looks great! Thanks for contributing this documentation.

PR #923, opened 2023-10-11
more accurate JupyterLab docs

dtenenba:
also documented how to ensure basic modules
like scipy and numpy are available in the JupyterLab session


dtenenba:
Let's de-emphasize jupyter notebooks more in favor of labs. And remove stuff like "the 2020 version is brand new".

PR #926, opened 2023-10-16
domain change mydb.fhcrc too mydb.fredhutch

fizwit:
Updating MYDB docs that use "fhcrc.org", changing to fredhutch.org.

Also changing the NGIX rewrite rules for mydb. All fhcrc.org will be rewritten to fredhutch.og


dtenenba (approved):
Thanks!

PR #934, opened 2023-12-06
remove tensorflow demo

atombaby:
The tensorflow demo was necessary back when we didn't have a module for it.  Now we have a mature module, so that's the preferred (supported?) method for using Tensorflow.


atombaby:
- [ ] Fix link at the end of this page: https://sciwiki.fredhutch.org/scicomputing/compute_overview/
 - [ ]

PR #946, opened 2024-03-01
updating workflow diagram in proof.md

laderast:
I fixed "Start a Cromwell Server" to "Start a PROOF Server"


tefirman (approved):
Good catch, @laderast , thank for fixing that!

tefirman (changes requested):
Looks like your branch is behind by a few commits compared to main, probably what's causing the merge conflict.

PR #952, opened 2024-03-11
update ssh config instructions on Win

chrisequalsdev:
Improved documentation on configuring an ssh config file on windows by linking to existing subsection. Also provides a basic config that will make `ssh rhino` work. This also will be easy to update the list of MACs when Jammy hosts start to become available.


vortexing (approved):
LGTM

PR #971, opened 2024-04-19
Add PHI page

### LINKED ISSUE #903: Link to de-id resource page
**URL of page(s) of concern**
https://sciwiki.fredhutch.org/datascience/deidentification/ does not actually link any where that I see easily to the resource library entry about it.

monicagerber:
This page provides guidance on different types of data sets that may contain PHI or are derived from healthcare data.

It will update and replace this page: https://extranet.fredhutch.org/en/u/consortium-clinical-data-access/data.html#coded

Needs to be reviewed by Jenny before reviewed here. 

Also fixes #903

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -84,6 +84,8 @@ datascience:
         url: /datascience/privacy_security/
       - title: Compliance and Legal
         url: /datascience/compliance_legal/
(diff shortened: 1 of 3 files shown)


vortexing (approved):
Full support.  Love this.

PR #972, opened 2024-04-19
Move clinical data page and update links

monicagerber:


vortexing (approved):
All wiring looks good.

Issue #439, opened 2019-11-07
Refresh content:  LabArchives Electronic Lab Notebooks


vortexing:
https://sciwiki.fredhutch.org/generation/clsp_labCov/#laboratory-data-management

@anupulletikurti is this info still accurate?  Would you like to update anything?

PR #980, opened 2024-05-10
overhaul and decommission data generation section

vortexing:
remove some legacy pages which are not required anymore and remove likely outdated material with no owner.  

Closes issue #439


vortexing:
Since my legacy self is the only person here to review this, I'm reviewing my own PR and approving it after checking that the preview built.

PR #984, opened 2024-05-14
quite the overhaul.

vortexing:
I'm working through the remaining decommisioning of the Data Generation section.  Now the last things that are remaining before full decom is the biospecimen banking which is being addressed in another issue.


monicagerber (approved):
Looks good! Love the clean up of old pages. Does any of the deleted content need to be re-written into the remaining pages?

monicagerber (approved):

PR #986, opened 2024-05-21
Add summary for temp storage overview

atombaby:
Tie the longer docs into the task-storage overview


_scicomputing/store_task.md
```diff
@@ -32,6 +32,20 @@ The Data Science Lab and Scientific Computing teams are available to help you id
 
 ## Storage Options for Tasks
 
+### Temp
+
+The [_Temp ](/scicomputing/store_temp) file system is available on all rhino/gizmo compute nodes.  It is a high-performance storage system mounted by rhino/gizmo nodes on the path `/hpc/temp`.  This system is not backed up, but snapshots are available.  Snapshots are available- there are two snapshots made every half-hour (meaning you can go back to a copy from up to an hour ago).
```
tefirman, line 37: Looks like you say "snapshots are available" twice in a row, maybe condense that into one, something like:

"This system is not backed up, but snapshots are available -- there are two snapshots made every half-hour..."

tefirman (approved):
Very minor change just for readability. Looks awesome otherwise!

PR #1007, opened 2024-05-30
Proof release notes for May v1.0.0

abbywall:
Next set of PROOF release notes


seankross (commented):
Should the release notes be according to date or should they be according to the release version itself? (Of course they should include the date)

abbywall:
@seankross We were thinking of this as news-like release notes - technical versioned release notes would be found in the repo

laderast (approved):
Approved.

PR #1011, opened 2024-06-07
WDL 101 -> Developing WDL Workflows

sitapriyamoorthi:
Updated developing WDL workflows guide content.


caalo:
Expanded changes so that we are consistent of using "Developing WDL Workflows" instead of WDL 101. Updated links also.

laderast (approved):

PR #1018, opened 2024-07-11
include smb (mac) and UNC (windows) paths to /hpc/temp

dtenenba:


atombaby (approved):
Added same verbiage to the temp docs in the scicomp section

Issue #1014, opened 2024-07-02
SSH connection from Windows requires OpenSSH client if not using Putty


chrisequalsdev:
### URL of existing content (a rendered, sciwiki.fredhutch.org link)

https://sciwiki.fredhutch.org/scicomputing/access_methods/#command-prompt-cmdexe

### Specific information requests or concerns, or specific region on page of content to refresh.

SSH connection from Windows also requires installation of OpenSSH client which can only be done by an admin. OpenSSH is also required to use the VSCode Remote-SSH extension.

PR #1023, opened 2024-08-05
add step for installing openssh-client + reorg

chrisequalsdev:
- Adds step to e-mail the Service Desk if ssh command is not available. 
- Reorganize all additional steps under one heading

Resolves https://github.com/FredHutch/wiki/issues/1014


laderast (approved):
Looks good. Glad we're mentioning the OpenSSH client on windows

PR #1029, opened 2024-08-13
update the Mac Finder path to /hpc/temp

dtenenba:
We have had two tickets from users for whom the original Mac Finder path that we had documented (`smb://hpc.chromium.fhcrc.org/temp.chromium`) did not work. Changing the path to `smb://fhcrc.org;USERNAME@hpc.chromium.fhcrc.org/temp.chromium` worked in both cases, so I figured we should document the path that works.

I am not sure why the original path stopped working and why the new one works (@atombaby can weigh in on that).

I'm also not sure if Windows users will run into the same issue, and if so, what the corresponding UNC path would be. If you know, can you add it to this PR?


chrisequalsdev (approved):
Using `fhcrc.org;USERNAME` is a workaround for a commonly encountered credentials issue (or something like that) on macs so I always have people use this by default. It's also documented that way in store_posix.md. I added HPC Temp to that page as well. This isn't required on Windows. I tested on Windows and our documentation is correct.

PR #1030, opened 2024-08-14
Fix order of arguments in S3 help command

jyelland:
The previous command had arguments reversed;
`aws s3 help cp` returns an error.


dtenenba (approved):
Good catch!

PR #1045, opened 2024-09-21
Update README.md: Update In-text Images

sitapriyamoorthi:
Update the example for including In-text images to include an underscore in the folder name. Figure does not render if it is not included.


sitapriyamoorthi:
Adding updates in a new draft pull request

PR #1090, opened 2024-11-15
Clarify data classification and scope for data sharing tools

atombaby:
Add some links and additional cautions for the reader to ensure that appropriate agreements are in place.  Recommend dataprotection for advice on transfers.


_scicomputing/hutchgo_overview.md
```diff
@@ -8,6 +8,9 @@ HutchGO- Hutch Globus Online- is a deployment of the Globus data transfer servic
 
 Globus is a non-profit service of the University of Chicago.  It is a combination of software (the Globus server and client tools) and services (the Globus control services) providing a number of different functions useful in research- data transfer, compute, and automation.
 
+We have a number of agreements in place with the Globus team that make it the preferred way to transfer sensitive data but it is essential that you are confident that transferring data with Globus won't violate any terms of use- contact `dataprotection@fredhutch.org` for advice regarding your data transfer.
```
tefirman, line 11:
```suggestion
Globus is the preferred way to transfer sensitive data, but it is essential that you are confident that transferring data with Globus won't violate any terms of use - contact `dataprotection@fredhutch.org` for advice regarding your data transfer.
```

_scicomputing/store_collaboration.md
```diff
@@ -3,32 +3,36 @@ title: Data Transfer and Storage Tools for Collaborating
 primary_reviewers: vortexing
 ---
 
-## Data Transfer Tools
+Before transferring data, please make sure you have appropriate protections in place- make sure you are allowed to share data via your desired mechanism, that agreements are in place or won't be violated by transferring data.  Contact `dataprotection@fredhutch.org` if you have questions or need assistance evaluating your transfer needs.
```
tefirman, line 6:
```suggestion
Before transferring data, please make sure you have appropriate protections in place and are allowed to share data via your desired mechanism. Contact `dataprotection@fredhutch.org` if you have questions or need assistance evaluating your transfer needs.
```

tefirman (changes requested):
Got feedback from Ty Lambert, 100% my fault on the delay. 

"For the data protection call out, I don't think it is necessary to discuss the agreements, it just feels off and like we are saying other partners don't have the same agreements in place (ie Microsoft or Amazon) which I don't necessarily think is true. If Globus is the preferred method of transmission, we should just state that, not that it provides any more protections than other methods. I haven't seen any of these agreements, but I doubt that this is the case. I would suggest changing this to reflect what is listed on the other page as a call out for data protections."

I've tried to add suggestions to reflect Ty's thoughts, but let me know if I missed something!

tefirman (approved):

PR #1129, opened 2025-02-05
update AlphaFold 3 docs

fizwit:
model and parameter data should be bind mounted

### DIFF
--- a/_compdemos/alphafold3.md
+++ b/_compdemos/alphafold3.md
@@ -63,6 +63,7 @@ mkdir -p $OUTPUT_DIR
 apptainer exec \
   --nv \
   --bind $DOWNLOAD_DIR:/root/public_databases \
+  --bind $MODEL_PATH=/root/models \
   $SIF \
   python /app/alphafold/run_alphafold.py \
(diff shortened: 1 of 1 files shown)


dtenenba:
This is not needed because we already have 

```
--model_dir=$MODEL_PATH
```

and 

`$MODEL_PATH` is `/root/public_databases/parameter_models/`.

Also, the syntax is wrong on the line you added, the `=` should be a `:`, but again, this is not needed.

fizwit:
lets use the real path to the parameter data:  export MODEL_PATH=/shared/biodata/alphafold3/parameter_models

fixed 
  --bind $MODEL_PATH:/root/models \

dtenenba:
I tried this and got:

```
FileNotFoundError: [Errno 2] No such file or directory: '/shared/biodata/alphafold3/parameter_models'
```

Inside the container /shared/biodata is not visible. 


I ran the existing [script](https://sciwiki.fredhutch.org/compdemos/alphafold3/) and it worked. I just changed BASE to a directory that had an input json file and an output/ directory.

Issue #1095, opened 2024-12-03
Link does not work for Mountain Duck macOS license


sitapriyamoorthi:
URL: https://sciwiki.fredhutch.org/compdemos/Mountain-CyberDuck/#mac-osx
This link is broken: https://teams.fhcrc.org/sites/citwiki/SciComp/Documents/fredhutch50.mountainducklicense

Page is not reachable to download the license

laderast:
Will be on self-service soon - what is timeline?

dtenenba:
Just pinged on the open helpdesk ticket, will update when I have a response.

laderast:
Addressed in PR: https://github.com/FredHutch/wiki/pull/1150

Issue #1106, opened 2025-01-09
Mountain Duck license file link broken 


pattygalipeau:
[**URL of page(s) of concern**
[Specify the URLs of the page(s) where there are technical problems](https://sciwiki.fredhutch.org/compdemos/Mountain-CyberDuck/)
https://sciwiki.fredhutch.org/compdemos/Mountain-CyberDuck/
license file, download the fredhutch50.mountainducklicense from here
the link is broken and doesn't take you to where you can download the license file

laderast:
Addressed in https://github.com/FredHutch/wiki/pull/1150

PR #1150, opened 2025-02-21
document installing mountain duck from Self Service on mac

dtenenba:
Also remove Windows 7 instructions as they are outdated.

This page still references the old way of getting S3 credentials but that needs to be fixed in a separate PR.


laderast (approved):

laderast:
Closes #1095 and #1106

PR #1161, opened 2025-03-14
Add announcement for maintenance

atombaby:


dtenenba:
If a job's run time will run into the maintenance window, will it be PENDING until after the maintenance? If so we should probably be explicit about that.

chrisequalsdev:
We need to reboot the rhino nodes (rhino01, rhino02 & rhino03)

Issue #1155, opened 2025-02-24
Add instructions for using Cyberduck with s3 using sso


chrisequalsdev:
### URL of existing content (a rendered, sciwiki.fredhutch.org link)

https://sciwiki.fredhutch.org/compdemos/Mountain-CyberDuck/

### Specific information requests or concerns, or specific region on page of content to refresh.

Current instructions assume use of keys. Keys are becoming rare in favor of sso. Update / add instructions for sso.

chrisequalsdev:
Notes:

https://docs.cyberduck.io/protocols/s3/#read-credentials-from-aws-credentials
aws configure
aws configure set aws_session_token

PR #1180, opened 2025-04-01
Document use of *duck apps with AWS SSO creds

chrisequalsdev:
Resolves: https://github.com/FredHutch/wiki/issues/1155


dtenenba (approved):
Thanks for doing this. Looks good.

Issue #1194, opened 2025-04-26
Add LLM Tuning/Training Guidance


vortexing:
Add into Data Science a section on AI use and the first item will be LLM tuning/training guidance.

PR #1200, opened 2025-04-27
AI for code development and LLM tuning

vortexing:
An initial draft of high level guidance.  Input welcome!  Specific resources people know about that help people learn to use AI for code well would be helpful but please only those that appear that they will "age well".


vortexing:
OK.  I also put the LLM tuning guidance in the same PR.  Closes #1200  and Closes #1194 .

PR #1209, opened 2025-05-16
update file name and path

chrisequalsdev:
In [9809576](https://github.com/FredHutch/easybuild-life-sciences/commit/98095764ebf40680e17a3a6cce13c1df2085fcbb) of the easybuild repo, a file used by the GitLab pipeline that builds the wiki was moved and renamed. This change should fix the build issue.


tefirman (approved):
Thanks for catching this, @chrisequalsdev !

PR #1211, opened 2025-06-02
Update references to miniconda with miniforge (preferred conda provider)

atombaby:
Update references from miniconda to miniforge.  Miniforge is the preferred and recommended conda provider- while miniconda has a less restrictive license than anaconda, it is still an Anaconda license whereas Miniforge is BSD-3


fizwit (approved):
LGTM

PR #1256, opened 2025-09-16
Some suggestions

atombaby:
- link from main proof page
 - some formatting- i'd really like folks to grok that it's "stop existing, start anew"


seankross (approved):
Looks great, thank you!

Issue #1225, opened 2025-07-18
Update pages that mention scratch storage


chrisequalsdev:
### URL of existing content (a rendered, sciwiki.fredhutch.org link)

01:
- [x] https://sciwiki.fredhutch.org/scicomputing/store_posix/#how-to-access-fred-hutch-storage
- [x] https://sciwiki.fredhutch.org/scicomputing/store_task/#scratch
- [x] https://sciwiki.fredhutch.org/scicomputing/store_posix/
- [x] https://sciwiki.fredhutch.org/scicomputing/store_overview/
- [x] https://sciwiki.fredhutch.org/compdemos/howtoRhino/
- [x] https://sciwiki.fredhutch.org/scicomputing/store_scratch/

02:
- [x] https://sciwiki.fredhutch.org/pathways/path-migrating-data-from-fast-to-cloud/
- [x] https://sciwiki.fredhutch.org/compdemos/motuz/#copying-files-to-cloud-locations
- [x] https://sciwiki.fredhutch.org/compdemos/globus-personal/
- [x] https://sciwiki.fredhutch.org/scicomputing/compute_cloud/
- [x] https://sciwiki.fredhutch.org/scicomputing/compute_restart/
- [x] https://sciwiki.fredhutch.org/scicomputing/hutchgo_overview/

03:
- [x] https://sciwiki.fredhutch.org/compdemos/ingest-Large-Data/
- [x] https://sciwiki.fredhutch.org/compdemos/Checkpointing-Gizmo/#a-more-realistic-example-with-local-scratch

PR #1261, opened 2025-09-19
Remove remaining scratch references

atombaby:
Updated last couple items from #1225 .  Removed checkpointing article as that is no longer applicable.


chrisequalsdev:
I'll merge this as I just spend a bit of time looking it over as well.

Issue #770, opened 2022-07-29
Consider integrating the AWS Lambda commit into gitlab


vortexing:
Right now last modified dates are continuously updated vi an automated commit after a AWS Lamba process.  We should consider sometime integrating that into the newer GitLab build process so that we have fewer moving parts.

dtenenba:
Good idea, especially as we need to get rid of the AWS account where that lambda function runs. We should bear in mind though that this process will still create new commits.

laderast:
@dtenenba is this already done? Can we close?

dtenenba:
Not done. Thanks for the reminder.

PR #1296, opened 2025-12-15
fixing last modified

seankross:
Spot checking this with a dozen individual pages, it seems like it works.


sitapriyamoorthi:
@seankross might be relevant: https://github.com/FredHutch/wiki/issues/770

PR #1300, opened 2025-12-19
Sitapriyamoorthi patch 1

sitapriyamoorthi:
Updating cBioPortal content so it tells the user that the mazimum foldxer upload size is 4gb :(


tefirman (approved):
Great call on adding this, just a couple small formatting tweaks. Thanks @sitapriyamoorthi !

PR #1304, opened 2025-12-24
add announcement for hpc temp config change

chrisequalsdev:


dtenenba (approved):
A relative link would be better (i.e. remove `https://sciwiki.fredhutch.org`, but no big deal.

PR #1318, opened 2026-01-27
Feat/cloud access docs

mbullmanFHCRC:


dtenenba (approved):
This looks good - it makes sense to move some of that info to CenterNet.

I notice that CenterNet page has some redundant documentation about Motuz that the wiki also has, but that's beyond the scope of this PR.

PR #1325, opened 2026-02-18
Working clarification

atombaby:
Clarify what entities are eligible for working directories


bmcgough:
I merged up from base (I hope that was the right thing to do). Otherwise, this is an improvement. Thanks.

PR #1326, opened 2026-02-18
remove manual last-modified-dates from yaml frontmatter

### LINKED ISSUE #1125: 'Last updated' information is inaccurate on wiki pages
**URL of page(s) of concern**
As an example: https://sciwiki.fredhutch.org/pathways/path-batch-computing/

**Describe the problem**
The page above according to github history was last updated 2024 but the page still says it was last updated 2022. Not sure if this is true of every single page but it's true on multiple pages. Here's another example: https://github.com/FredHutch/wiki/commits/main/_scicomputing/store_posix.md it says it was last updated 2023, actually last updated 2024. 

It's not a huge problem but it would be nice to show off how well maintained some of these pages actually are.

dtenenba:
This ensures that the generated/displayed dates do not get out of date.

Closes #1125 . See my [comment](https://github.com/FredHutch/wiki/issues/1125#issuecomment-3918297749) there for details.


dtenenba:
The artifact of this built PR is [here](https://wiki-artifact-app.fredhutch.org/artifacts/50142) if you want to check the dates....

atombaby (approved):

PR #1358, opened 2026-04-10
Update collaboration folder info

atombaby:
Reorg and refine information, add restriction on where we're putting these directories


_scicomputing/access_permissions.md
chrisequalsdev, line 1: This is confusing to me "Each PI has a default group, typically (though not exclusively) lab members."

chrisequalsdev (approved):

Issue #1352, opened 2026-03-24
Unclear preferred $TMPDIR for Gizmo jobs


derrik-gratz:
### URL of existing content (a rendered, sciwiki.fredhutch.org link)

https://sciwiki.fredhutch.org/scicompannounce/2024-04-02-new-hpc-temp-storage/

https://sciwiki.fredhutch.org/compdemos/store_job_local/

### Specific information requests or concerns, or specific region on page of content to refresh.

Between these two pages, it's not clear what a user should use for TMPDIR space on large jobs. I was using the default assigned $TMPDIR (`/local/scratch/...`) and encountering job failures. Per @tefirman , `/hpc/tmp` is preferred for large scratch space. 

### @ GitHub username of the listed primary reviewer of the page if known

@atombaby

PR #1359, opened 2026-04-22
Down the temporary file rabbit hole

atombaby:
Expanded documentation of temporary files and temp data management:
 - More info about storage options
 - Longer discussion around TMPDIR and configuring application temporary files
 - Reorg of some files from demos into _scicomputing

See also #1352


dtenenba (approved):
This looks great to me.

PR #1368, opened 2026-06-02
specify --use-device-code with aws sso login and aws configure sso

dtenenba:
This is another thing that is seemingly required with the new AWS CLI module, we have gotten several tickets from people getting errors because they did not include this flag. 
Would be great to get a speedy approval on this. Thanks!

### DIFF
--- a/_compdemos/aws-s3.md
+++ b/_compdemos/aws-s3.md
@@ -263,7 +263,7 @@ Note that the `aws sso login` step will require you to copy a URL into your brow
 ```
 ml purge
 ml awscli
-aws sso login
+aws sso login --use-device-flag
 ml fhR
 R
 ```
--- a/_compdemos/aws-s3.md
+++ b/_compdemos/aws-s3.md
@@ -263,7 +263,7 @@ Note that the `aws sso login` step will require you to copy a URL into your brow
 ```
 ml purge
 ml awscli
-aws sso login --use-device-code
+aws sso login --use-device-code # omit this flag when using older versions of the AWS CLI
 ml fhR
 R
 ```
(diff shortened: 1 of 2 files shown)


atombaby:
I can't find any documentation for `--use-device-flag`.  Is this `--use-device-code`?  Do we know what caused this to break?

dtenenba:
You're right, will fix.

I built a more recent AWS CLI module upon user request and it's now the default. It assumes the user is running "aws" on their local machine and spins up a web server bound to localhost. Since most of our users are on a rhino/gizmo system that won't work. 

`--use-device-code` gives you the old behavior where no server is spun up on localhost; instead you go to an AWS server and enter the code displayed.

atombaby:
Does `--use-device-code` work in the same way for earlier versions of the CLI?

dtenenba:
fixed now, great catch

dtenenba:
> Does --use-device-code work in the same way for earlier versions of the CLI?

Argh, no, it's unknown in the next earlier version. 

IDK why they need to change the default behavior like this. Just to spin up a spyware server on the local host I guess.

Not sure what we should do here....either specify that this flag is required with version >= 2.32.6 or just document the commands without the flag and indicate the flag should be added if the user gets an error with 127.0.0.1 in it?

atombaby:
Yeah, that's a tough one.  Thinking for the future environment we bake this into the OS so we can manage releases better, but for now I'd guess that most folks use the default, which is latest, so make the primary documentation match that.  Or maybe a roll-up?

<details>
<summary>AWS CLI version latest</summary>
Use `aws sso configure --use-device-code ...`
</details>
<details>
<summary>AWS CLI older versions</summary>
Use `aws sso configure ...`
</details>

dtenenba:
Roll-up is a good idea but doesn't work so well in context where the flag goes/doesn't go in one command of many. So I just opted for some awkward text indicating the flag should be omitted for older versions of the AWS CLI.

atombaby (approved):

PR #1370, opened 2026-06-11
Remove sentence plugging iterm2

emjbishop:
The built-in terminal should be sufficient for this Pathway. Also, I don't see how this tool or its developers are connected to Fred Hutch so I don't think we should request donations to them from our readers.


atombaby (approved):
Yeah, it's a little tacky.  If we're that chuffed about iterm2 we probably should look at an institutional contribution.

